PR #45, opened 2026-01-13
Power

Raikyr0:
## About the PR
Some power related changes.

## Why / Balance
So, this PR increases the power storage of SMESs by around 15 times.
The reasoning given by EE at the time was it emulates SS13's SMES values as actual back up power, as currently the 3 full SMES at round start empty in about 5-10 minutes. This change will make it so it will last substantially longer, and thus remove the need to break into engineering early on due to the lack of staff. Long enough for head of personnel to actually HIRE someone over the need to break in. This also increases the max sell value of SMES so engineers can sell power more reliably. This also makes it less likely a power sink will be noticed by an immediate power outage.
I've also noticed that Delta V doesn't appear to have up-gradable machines. We might have to discuss that at a later point.
Increased basic glass Solar power generation by 33% and upgraded panels by 100%,
Solar power has always felt inconsequential, and the act to upgrade them is outright tedium with little overall impact.

TLDR; Power Storage Up = Backup Power stays on/SMES sells for more.

## TO DO
So we still need to discuss what we want to do about DV's exploding TEG chamber, since that's supposed to create busy work for Engineering by forcing alternative power sources, but that sacrifices opportunities to role play. This will also mean all the current maps need to be reworked to survive TEG automatically exploding, so it may be simpler to remove it then ask map makers to redo all the TEGs.

**Changelog**
:cl:
- tweak: Increased SMES power storage.
- tweak: Increased Solar power generation.
- tweak: Placeholder Line


SyaoranFox:
This PR is something new, so I am going to recommend we look at this later, after all the base features are in.

PR #46, opened 2026-01-14
Wooden Railings Port

Dunrab:
Brought over the wooden railings that are used for some maps and shuttles.

## About the PR
We used these wood railings on some of our maps and shuttles so they will be needed.

## Why / Balance
These sprites were used in our maps and by players in round

## Technical details
Had to rename BaseRailing to BaseRailing1 since delta-v has their own railings that uses the same id. Should also note since the camo walls were in the same folders as the wooden railings (for better or worse) I chose to bring those over in this PR as well. I can always take them out and put them in a separate pr if needed.

## Media
Small image of the railings in game 
wood railing

## Requirements

**Changelog**

:cl:
- add: Ported Floofs wood railings mapping

### DIFF
--- a/Resources/Prototypes/_Floof/Recipes/Construction/structures.yml
+++ b/Resources/Prototypes/_Floof/Recipes/Construction/structures.yml
@@ -6,10 +6,6 @@
   startNode: start
   targetNode: railingWood
   category: construction-category-structures
-  description: woodenrailing-descrption
-  icon:
-    sprite: _Floof/Structures/Walls/railing.rsi
-    state: wood_side
   objectType: Structure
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/_Floof/Recipes/Construction/Graphs/structures/railing.yml
```diff
@@ -0,0 +1,90 @@
+- type: constructionGraph
```
Mnemotechnician: This adds a construction graph, but no respective construction prototype?

Memeji (approved):

PR #50, opened 2026-01-16
Port Weather Cycles

Dunrab:
Port over the weather system @Mnemotechnician made for our planet side maps.

## About the PR
The weather system is a great addition to all our planet side maps allowing for random weather vents to take place on those maps that can lead to some variation of roleplay.

## Why / Balance
All our planet side maps use this feature and it would be nice to keep them with their random weather.

## Technical details
Had to change a few ftl files since the integrations tests did not like that their counter started at 0 instead of 1, additionally I had to change a prefix for one of the weather prototypes since weather-hot was used twice when it should have been weather-warm.

Note: This will need to be tested again once our maps are in to make sure the random weather events happen.
**_This will need testing_**

## Media

Will try to get a video posted for this, I need to tweak my screen recorder to try to reduce the file size enough to pot it to github.

## Requirements

**Changelog**

:cl:
- add: Ported the weather system for our plant maps

### DIFF
--- a/Content.Server/_Vulp/Weather/Functions/WeatherDamageMobs.cs
+++ b/Content.Server/_Vulp/Weather/Functions/WeatherDamageMobs.cs
@@ -4,6 +4,8 @@ using Content.Shared._Vulp.Weather;
 using Content.Shared.Alert;
 using Content.Shared.Body.Components;
 using Content.Shared.Damage;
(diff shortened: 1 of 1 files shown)


Dunrab:
i fat finger uploaded something again smh, hopefully that will pass all checks now

Mnemotechnician (approved):

PR #58, opened 2026-01-19
Ported Quads To Chassis Selection System

StrawOfMan:
## About the PR

Ported the 4 kinds of Quadborgs (Security, Service, Medical, Salvage) to function under the new chassis selection system. Does **not** include the quad endoskeleton or the CC Quad.

## Why / Balance

The people want Quadborgs.

## Technical details

Quads are now integrated into the new system chassis selection system. When a generic chassis is constructed and the controlling player calls up the UI, these 4 chassis will now appear as options within the list.
Codewise, they're nearly completely identical to their non-quad counterparts, with the exception of their visuals and having the capability to lay down.
Because of this, however, the quad endoskeleton has been made entirely redundant, as this new system doesn't really allow for multiple kinds of endoskeletons with unique chassis options. Trying to add this functionality is out of scope for myself.

## Media

Chassis-Selection

## Requirements

## Breaking changes

Quad endoskeletons are no more, they should be removed from any maps that have them.
Role slots for Medical and Service Quads are made redundant by this system, so that should be adjusted.

**Changelog**


Mnemotechnician (commented):
This looks like a good re-implementation. I'm assuming the modules the secborg receives are the standard delta-v security cyborg modules so the balancing issues from floofstation can be forgotten about.

Please fix the yml formatting in Resources/Prototypes/_Floof/borg_types.yml. Blank lines in the middle of block mappings make the prototypes borderline unreadable, especially considering how inconsistent their placement is.

Mnemotechnician (approved):
Test fail is caused by an upstream bug, I'll push a temporary fix soon

PR #59, opened 2026-01-20
Botany, Ice Cream, and Food

SyaoranFox:
## About the PR
Tried to add in all of the botany and food recipes that we were missing.

## Why / Balance
We had them before, and now we have them here.

## Technical details
A lot of yml entries with FTLs in their appropriate namespaces.

## Media
Its food my dudes, imagine a sandwich.

## Requirements

## Breaking changes
So, there is no support for multiple items being made from a single recipe. That would involve modifying how microwaves recipes work and I just didn't have it in me today. Humpshroom and Breedshroom have not been added yet. There are some recipes in here that are meant to have flavorol, but instead have wine. This was here when Delta V was developed.

**Changelog**

:cl:
- add: Added NF recipes
- add: Pomegranate seeds and fruit
- add: Flavorol
- add: Ice cream and ice cream recipes.

### DIFF
--- a/Resources/Prototypes/_NF/Entities/Mobs/NPCs/animals.yml
+++ b/Resources/Prototypes/_NF/Entities/Mobs/NPCs/animals.yml
@@ -1,0 +1,14 @@
+﻿- type: entity
+  name: bee
+  suffix: Angry, Timed
+  parent: [ MobAngryBee ]
+  id: MobAngryBeeTimed
+  components:
+  - type: MeleeWeapon
+    angle: 0
+    animation: WeaponArcBite
+    damage:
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/_Floof/Hydroponics.yml
```diff
@@ -0,0 +1,26 @@
+﻿- type: seed # Floofstation
```
Mnemotechnician: This file is certainly supposed to be Prototypes/_Floof/Hydroponics/seeds.yml instead of Prototypes/_Floof/Hydroponics.yml

PR #74, opened 2026-02-03
Epi

Raikyr0:
## About the PR
Porting the Artifact Changes people enjoyed.
Minor changes to hellspawn, it's now properly labeled as a lesser hellspawn, can now be dispelled for a chunk of damage, and speaks universal.
Added a lesser behonker to artifact spawns.

I haven't checked if the lewd toys spawner has been put in or not so it may initially fail.

Edit: Ok, I just saw DV giga updated their system, so the original format artifact effects won't work, and I also saw their gun effect is disabled because not working. So all we're getting is summoning demons. Supermatter may have not been ported yet so that might also fail.

**Changelog**
:cl:
- add: Ported the Floof Artifact effects that everyone loves
- tweak: changed somethings about the friend who shows up

### DIFF
--- a/Resources/Prototypes/_DV/GameRules/glimmer_events.yml
+++ b/Resources/Prototypes/_DV/GameRules/glimmer_events.yml
@@ -6,7 +6,7 @@
     - id: NoosphericZap
     - id: NoosphericFry
     - id: PsionicCatGotYourTongue
-    #- id: MassMindSwap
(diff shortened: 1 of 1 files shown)


SyaoranFox:
Took a look. It seems fine now, but please modify the two tests that fail. If we don't do it here, all other builds will fail the same test. Once that's done, I'll merge it.

PR #79, opened 2026-02-06
Lightweight and Heavyweight Traits

Cprn-Sprkl:
## About the PR

Re-implemented the Lightweight and Heavyweight traits.

## Why / Balance

To bring parity with Floof.

## Technical details

Trait prototypes and ftl name/desc entries for each. Trait effect to let traits alter entity fixtures.

## Media

image
image
image

A few tests show this working on the allowed species. Pictured example is a Tajaran's fixture being reduced by Lightweight and increased by Heavyweight.

</p></details>

## Requirements

## Licensing:

## Breaking changes

None known.

**Changelog**

:cl: sprkl
- add: Weight change traits restored!

### DIFF
--- a/Resources/Prototypes/_Floof/Traits/disabilities.yml
+++ b/Resources/Prototypes/_Floof/Traits/disabilities.yml
@@ -41,4 +41,4 @@
     - Lightweight
   effects:
   - !type:ModifyFixtureEffect
-    factor: 1.4 # Lets try the opposite of lightweight, yeah?
+    factor: 1.1832 # Square root of 1.4 as this affects the radius, opposite of lightweight.


Mnemotechnician (approved):
I'd like to point out that on EE floofstation these traits affect fixture density and not size, but I think this is the better option anyway as it makes the heavyweight trait have cons.

Mnemotechnician (changes requested):
Okay, actually there's one change needed. Since these traits now modify the fixture radius instead of density, they have quadratic effect on the mass of the entity since the mass is determined as `radius^2 * density * 2 * pi`, so these adjustments are needed (otherwise the lightweight trait would reduce your weight to 36% of the original while the lightweight one would increase it to 196%.

Mnemotechnician (approved):

PR #90, opened 2026-02-11
Fixed minor issue

JFerrix:
This is literally a three line change, I changed placeholders that I overlooked so now you can actually use colours for shadowkin, tajaran and resomi. Previously they were humantoned because I rebuilt the species definitions over those but I forgot to edit that so here that is

## Requirements

## Licensing:

**Changelog**
:cl:
- fix: Changed Tajaran, Shadowkin and Resomi to use hsv colors instead of skintones.


Mnemotechnician (approved):
I'm pretty sure kitsune need the same treatment

PR #101, opened 2026-02-15
Surgery Revert

Raikyr0:
## About the PR
Reverts Surgery

## Why / Balance
Removes the changes that make surgery unfun and tedious to the point people quit Funky

**Changelog**
:cl:
- tweak: Reverted to less tedious surgery.


SyaoranFox:
> I don't think I even need to explain this.

I think it should be explained. I was testing on the server with Marzi, and even as CMO, the surgery speed was absurdly slow.

PR #106, opened 2026-02-20
Loadouts

Dunrab:
## About the PR
Add hats, jumpsuits, backpacks, gloves, outer clothing and shoes to every job. Also fixed a problem I noticed with my previous PR where min amount for jumpsuits for an example was set to 0 when it should not have been and has been corrected to a min amount of 1 and a max amount of 1.

## Why / Balance
Allows for more character customization and we had it this way in the old code base.

## Technical details
Every job is parented from one of the following:

FloofHats, FloofJumpsuits, FloofBackpacks, FloofGloves, FloofOuterClothing, FloofShoes. This makes it easier for us to make changes to loadouts in the future and makes things much easier to read and or follow.

## Media
Will post a few examples below once this passes tests.

## Requirements

## Licensing:

**Changelog**

:cl:
- tweak: Tweaked loadouts to add jumpsuits, shoes etc. to each job.


Memeji (approved):
Testing looked good on my end from what I could see.

PR #122, opened 2026-02-23
The marking communism patch

JFerrix:
## About the PR
The best way I can describe this is as doing communism on markings. Basically this ports pretty much every marking I could find from floofstation, removes every single species restriction and merges every identically named marking file into one. Also moves all non-core directory folder markings into the floof customization folder. I could have moved the core ones aswell due to all of them being changed aswell but its the core folder.
This is basically it. This PR does just that but its a massive change and I anticipate upstream merges becoming slightly more difficult. Then again fixing this type of problem is now much easier and therefore faster than before so id say it cancels itself out.

## Technical details
I usually dont do this often but this time it IS required:
Merged basically everything that isnt in the base Resources/Prototypes/Entities/Mobs/Customization/Markings into the _Floof/Prototypes/Entities/Mobs/Customization/Markings folder.
Done due to two simple reasons a) makes it easier to maintain. Seriously I doubt I am the only one that gets an aneurism seeing vulpkanin markings scattered over three different codebase folders despite treating the same thing
b) literally everything that isnt the hair marking folder in the base folder got changed

## Requirements

## Licensing:

## Breaking changes
There is too much. This patch spans over 2000 files, its too much

**Changelog**
:cl:
- add: Added Floof markings
- remove: Removed every single species restriction
- tweak: Repacked marking files

### DIFF
--- a/Resources/Prototypes/Species/human.yml
+++ b/Resources/Prototypes/Species/human.yml
@@ -21,14 +21,20 @@
     TailBehind: MobHumanoidAnyMarking # Floofstation
     TailOversuit: MobHumanoidAnyMarking # Floofstation
     Head: MobHumanHead
+    Face: MobHumanoidAnyMarking # Floof
+    Snout: MobHumanoidAnyMarking # Floof
+    SnoutCover: MobHumanoidAnyMarking # Floof
     Hair: MobHumanoidAnyMarking
     FacialHair: MobHumanoidAnyMarking
-    Snout: MobHumanoidAnyMarking
     UndergarmentTop: MobHumanoidAnyMarking
     UndergarmentBottom: MobHumanoidAnyMarking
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
> Merged basically everything that isnt in the base Resources/Prototypes/Entities/Mobs/Customization/Markings into the _Floof/Prototypes/Entities/Mobs/Customization/Markings folder

This is terrible for maintenance. Anyone using an IDE has access to project-wide search (ctrl-shift-F) and project-wide search-and-replace (ctrl-shift-R), there's no real maintenance burden. What IS a maintenance burder though is having all of them clumped together in one single file, having upstream files modified and outright removed (welcoming merge conflicts with open arms), having markings that are implemented in different ways (see: vulp tails) clumped together with regular markings and and... just looking at this is going to give me an aneurism as the person usually performing upstream merges.

PR #125, opened 2026-02-25
Trait Costs

SyaoranFox:
## About the PR

Made the costs reflect the impact of each trait. Some add points, some take points away.

## Why / Balance
Some traits are purely beneficial, and thus should take away more than a single point. This is how Floof worked.

## Technical details
In our original system, we listed points rather than cost. Everytime we put a new trait in, we must ensure we list its cost. If the trait is negative only, the cost should be made negative to refund points back to the selector.

## Requirements

## Licensing:

**Changelog**
:cl:
- tweak: Costs of traits currently present.

### DIFF
--- a/Resources/Prototypes/_DV/Traits/medical.yml
+++ b/Resources/Prototypes/_DV/Traits/medical.yml
@@ -78,7 +78,7 @@
   name: trait-redshirt-name
   description: trait-redshirt-desc
   category: Medical
-  cost: -8 #floof
+  cost: -6 #floof
   effects:
   - !type:AddCompsEffect
     components:


Resources/Prototypes/_DV/Traits/medical.yml
```diff
@@ -73,6 +78,7 @@
   name: trait-redshirt-name
   description: trait-redshirt-desc
   category: Medical
+  cost: -8
```
Mnemotechnician: Redshirt simply removes the 100-200 dmg crit state, I don't think it's worth 8 points (if any at all) since being in crit is essentially the same as being dead (and we dont even have softcrit on panta-rhei).

Resources/Prototypes/_DV/Traits/mental.yml
```diff
@@ -20,6 +21,7 @@
   name: trait-depression-name
   description: trait-depression-desc
   category: Mental
+  cost: -6
```
Mnemotechnician: Why is depression worth the same as pacifist?

Mnemotechnician (changes requested):
Other than that it's fine, but please add marker comments

Resources/Prototypes/_DV/Traits/medical.yml
SyaoranFox: its to match what we had on Floof. From what I have heard from redshirt players, or salvagers who have tried redshirt, its like playing on hardcore mode.

Resources/Prototypes/_DV/Traits/mental.yml
SyaoranFox: Cause permanent walk slow unless you have your meds.

PR #130, opened 2026-02-26
trying to get old floof rules onto the new server

SyaoranFox:
I am just bringing the old rules over to allow us to have some kind of rules that match our server. Approved by Tech Coarse. I need help with it though, I am so tired.


SyaoranFox:
I can't figure out why it keeps throwing "proto: Reading Content.Shared.Guidebook.GuideEntryPrototype(FloofstationRuleset) threw the following exception: System.ArgumentException: No data definition found for type Robust.Shared.Prototypes.ProtoId`1[Content.Shared.Guidebook.GuideEntryPrototype] with node type Robust.Shared.Serialization.Markdown.Mapping.MappingDataNode when reading"

PR #147, opened 2026-02-28
Harpy wing layering bugfix

JFerrix:
## About the PR
Basically harpies had a default wing marking which didnt get overwritten causing it to not get removed and being rendered underneath any other wing.
Only solution is to remove the default marking, which means that harpies in character creation will not have any wings. They need to be manually added during the character creation as markings.
Does it look bad? Yes, is there another way to fix it? Not that I have found one

## Requirements

## Licensing:

**Changelog**

:cl:
- remove: Removed harpy default marking

### DIFF
--- a/Resources/Prototypes/_DV/Species/harpy.yml
+++ b/Resources/Prototypes/_DV/Species/harpy.yml
@@ -107,8 +107,8 @@
       required: false
     Arms:
       points: 10
-      required: false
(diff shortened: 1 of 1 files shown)


Mnemotechnician (commented):
This is reasonable I suppose, but why required: true?

Mnemotechnician (approved):
Actually the field required is unused, so it's whatever

PR #205, opened 2026-03-06
PTL Tweaks and Fixes

PhosphateWithManyOs:
## About the PR
Makes the ramping formula much more comprehensible. PTL will now at most generate 5k per shot but would require a significant amount of power to reach this.

## Why / Balance
The equation is modeled after capacitor charging in terms of power per speso. 
Also fixed the PTL only using 1kW of power. Removes the PTL from the cargo buy for now.

## Technical details

spesos = 5000*(1-e^{-.5x})

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- tweak: Adjusts the PTL


Mnemotechnician (approved):
Im not sure if its worth removing from cargo listing, but sure

PR #212, opened 2026-03-06
Coat sizes (and other balance adjustments)

JFerrix:
## About the PR
First of all fixed the captains armored wintercoat a) being based off the captains carapace and having 4x4 slot with the same armor and b) also nerfed the armor values of the longcoat to be better suited for the game and also to make the carapace worth taking instead of just being a style choice. The wintercoat now has a flat 30% damage reduction against blunt, slash, pierce, heat and caustic. It also lost its stamina defense because... if you want to be heavily armored, use the carapace.
Also buffed the captains hardsuit because it was basically a salvage hardsuit. I am prepared to die on that hill but a literal armored hardsuit should not give less protection than a basic security vest. Especially not if it also has a 20% slowdown. Its armor-stats are now much more comparable to the carapace being slightly worse against blunt damage but being substantially better against slash and pierce. Its also weaker against heat, decent against radiation and extremely effective against caustic. Dont think I ever actually saw someone die to caustic but well if its armored and covers every single area of your body, why wouldnt it be heavily resilient against superficial caustic burns?
All of this comes with an additional speed nerf, it now slows the wearer by 30% instead of 20%. I would legitimately like to delete the carapace or nerf that one substantially too but I feel that wold go to far. Lets see if this gives players a reason to actually wear and use the hardsuit other than "area depressurized, need space suit".
Update, I actually nerfed the carapace to have a 10% speed penalty and have slightly worse protection.
The general progression is supposed to be: Light armor without downsides, use the captains coat, more armor than most secoffs and the HOS: Carapace but now you cant run as fast anymore and: "Everything is going to shit, where is my budget jugsuit?"- hardsuit. I feel like this is a substantially better balance and most importantly, discourages captains from being both the most qualified and most effective person at the same time.
Captain always suffers from the fact that its both the most durable and best armed role while also having AA. Which often leads to captains not playing around the fact that they have a security team and instead yolo in on their own just because they can.
Not a healthy state to be in, I hope this changes that.

## Requirements

## Licensing:

:cl:
- tweak: Buffed captains hardsuit to be an actually viable choice
- tweak: Made captains wintercoat lighter armored for being a no slowdown armor with pockets
- tweak: Gave carapaces a slightly worse armor and a 10% slowdown

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
@@ -431,9 +431,9 @@
     modifiers:
       coefficients:
         Blunt: 0.6
-        Slash: 0.4
-        Piercing: 0.4
-        Heat: 0.6
+        Slash: 0.5
+        Piercing: 0.5
+        Heat: 0.7
         Radiation: 0.5
         Caustic: 0.3
   - type: DamageOnShootProtection # imp
@@ -444,8 +444,8 @@
   - type: StaminaResistance # DeltaV
     damageCoefficient: 0.3
   - type: ClothingSpeedModifier
-    walkModifier: 0.8
-    sprintModifier: 0.8
+    walkModifier: 0.75
+    sprintModifier: 0.75
   - type: HeldSpeedModifier
   - type: ToggleableClothing
     clothingPrototype: ClothingHeadHelmetHardsuitCap


JFerrix:
This were the intended stats originally, I wanted to have it be more vulnerable to blunt than to pierce mostly to make it strong against bullets but if someone walks up to you and beats the shit out of you, you lose. I also wanted to movespeed to be 30% slow originally but I thing 25% is a good middle ground. I dont want to nerf this to be much worse than the carapace was before but it should replace the carapace in regards to its performance while the carapace is meant to be slightly worse but with less slowdown

PR #229, opened 2026-03-07
Remote cwoinks

Myzumi:
## About the PR

Allows Remote Sending Cwoink Messages to Users.

## Why / Balance

No Online Requirement anymore.

## Media
<details><summary><h3>Click to show</h3></summary> <p>
image
</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/_DV/Curation/Systems/CwoinkSystem.cs
+++ b/Content.Server/_DV/Curation/Systems/CwoinkSystem.cs
@@ -553,15 +553,27 @@ public sealed partial class CwoinkSystem : SharedCwoinkSystem
         // Otherwise patch (edit) it
         if (existingEmbed.Id == null)
         {
-            var request = await _httpClient.PostAsync($"{_webhookUrl}?wait=true",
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
If I understand the code correctly, this just adds a new API endpoint that allows sending a cwoink over HTTP? Is this intended for personal use, or is there more to it?

SyaoranFox:
I do have concerns around what happens if many cwoinks are sent but the user never returns. Will these messages need to be stored forever? Over many years, many users, many cwoinks, could this lead to server bloat/slowdown? Happy to see this go ahead, these are just thoughts I had when I saw the implementation.

Would recommend some way of doing backlog clean-up, or a mechanism to clear pending messages.

Mnemotechnician:
@SyaoranFox this has no effect on how messages are sent to users, it only changes how the system goes about relaying them to discord (discord webhook API calls can randomly fail either due to webhook rate limiting or for no reason whatsoever)

PR #267, opened 2026-03-12
Coyote genital markings V1

JFerrix:
## About the PR
This PR adds coyote genital markings to all species with a default amount of 10 points and incorporating the floof undie system. That was players can remove and change those during the game.
Worth noting, this is an implementation that does not use the consent system, meaning that everyone will see said markings. I have no idea if this is a good idea but I lack the knowledge or understanding to implement a solution that works for the consent system.

Update: Fixed the licensing issues, gotta figure out solutions to these two issues. No idea how yet

ToDos:
- [ ] Find a way to hide markings when clothes are on
- [ ] Find a way to add a consent toggle

## Requirements

## Licensing:

**Changelog**
:cl:
- add: Added coyote genital markings
- tweak: Rearranged visual layers for species to prevent clipping issues
- tweak: added genital layers to the undies system


Mnemotechnician (changes requested):
Regardless of license, this needs a consent toggle, otherwise these markings are gonna be a bait for accidental consent breaches. ESPECIALLY the huge markings that WILL show up under clothes.

Also, there is a lot of undocumented changes to upstream files.

jimmy12or:
It may take some snooping around, but I know Hardlight server has the genital markings working with client side enable/disable settings.

PR #278, opened 2026-03-15
Sorting traits

Raikyr0:
## About the PR
Changes up the categories for traits, everything fitting under medical didn't feel right.
This PR splits Disability into Disability Physical and Disability Mental
Adds Cybernetics, Psionics and Lewd categories
replaces the Mental category with Feats - the DnD term

Most things that were under medical are in in the disabilities tabs now.
There's also a skills category that we have but it's effectively the same as feats.

## Why
I figured, it's better to sort it earlier then later.

**Changelog**
:cl:
- tweak: Sorts the Traits into new categories.


Mnemotechnician (commented):
Overall this looks fine, but changes to upstream files need to be commented.

Cprn-Sprkl:
While we're changing the category names it might be a good idea to drop "disability" from the names completely. Just have Physical and Mental alongside the other categories. Especially since the old Mental category got replaced completely.

Mnemotechnician (approved):

PR #286, opened 2026-03-16
Max Language increase

Raikyr0:
## About the PR
Increases max number of languages taken from 2 to 5
They currently cost 1 point each so it would cost 3 more points for 3 more languages.

**Changelog**
:cl:
- tweak: Increased max number of languages taken from 2 to 5


Mnemotechnician:
Obsolete

PR #294, opened 2026-03-20
Vector change

OrenjiKnight:
## About the PR

Brought over the vector from the old base, however changed enough of it so that it feels like its own unique weapon.

## Why / Balance

The standard vector and kriss vector are both chambered in .45 magnum carrying 40 round clips, however A: the kriss vector is almost exclusively to admins so shouldn't see that much game, and the standard one for security:
- Costs 67k to buy one
- the magnum magazines were changed to be a 1x3 spot in the inventory rather than a 1x2
- It is stuck to a two round burst
The vector for the syndicate has a faster firing rate than any other firearm in the game (besides the kriss vector) but it lacks a lot of accuracy. It also is chambered in an expanded .10 chamber because of its firerate. Standard magazine holds 24 shots which all do about 5 piercing which means that one magazine can kill just about one person per. The gun itself costs about 12 TC to purchase off the uplink and the standard magazine is worth 1 TC. .10 was also expanded on due to how lacking the ammo was (only used as a projectile for the minigun) other ammo types were added, which include:
_- Standard_ (copy of original .10): 5 Piercing
_- Rubber:_ 2 blunt
_- Practice:_ 1 blunt
_- Uranium:_ 2 piercing, 3 radiation
_- Armor Piercing:_ 2 piercing (due to damage, you're not going to do more than 48-50 total damage to someone)
_- Hollow-point_ (specialty made): 6 slash, 1 piercing
All of the specialty magazines cost to 2-3 TC per pop, and still should only do enough damage to kill one person per.

Back to the accuracy, one-handed, all of the weapons have horrible accuracy past a meter. When wielded, the two full vectors (gen2 and the kriss) have better accuracy for an SMG in burst, and the modded vector has decent accuracy but for how fast it shoots, it still does not do as much damage as the others past a few meters.

## Technical details

The most that changed is mostly the ymls for SMGs and the according ammunition for them as I added the ammo-type for the new weapon.

## Media

https://github.com/user-attachments/assets/a3b02169-3680-40a7-8381-2b543b6ee3dd

</p></details>

## Requirements

## Licensing:

## Breaking changes

I do not believe there should be any.

**Changelog**

:cl: "OrangeKnight"
- add: Vector 
- add: Modded Vector (Syndicate)
- add: Unique Ammo for Modded Vector
- add: Kriss Vector (Admeme)
- tweak: Vectors can be wielded
- tweak: Changed how the weapons work functionally.
- tweak: Logi store & syndicate uplink have respective weaponry


OrenjiKnight:
> atp if your gonna bring that gun back to the crew at least change the name so our asses dont get sued and ditch the 2 bullets per shot (possibly chamber it to 35 auto and possibly give it to hos?)

1: okay, change the name to something else, sure
2: The whole point of bringing it over was to make it functionally unique to all of the other security SMG options already, I mean you have:

The WT:
 - One handed
 - Full auto
 - low accuracy

The Drozd:
 - Basically the old vector, just newer

And The C-20
 - On par with the Drozd more less

Then there's also the Patos, but that's more or less a machine-pistol

The whole idea is for the gun to be something _new_ while still being balanced enough to not be the first choice. Also, its the only thing that can actually use the .45 magnum magazines that are coded in because they were brought over without the actual gun related being brought over. The idea was to make as close as a replica as possible without making it a copy because the actual vector is a beast, that's why it's chambered how it is, that's why it shoots how it is. There was a lot of thought actually put into balancing it so that it felt fun to play with without sacrificing the usefulness of the other weapons.

3: I mean, It would be a pretty baller head of security weapon, if the head of security loadout system worked like that right now. The HOS gets the current choice between a X-01 _(I believe its called)_ and the energy magnum. Both of which would be immediately out-shadowed by the gun due to it being an actual ballistic longarm that is not the eshotty. _(See old floof hos work and everyone rocked either the c-20, bulldog, or wt.)_

Mnemotechnician (changes requested):
All new prototypes need be moved under the _Floof namespace.
<h6>also 67 joke is kill, at least make it 69</h6>

Mnemotechnician (approved):
I suppose this is fine

PR #314, opened 2026-03-25
Sex Trait - Honey

Raikyr0:
## About the PR
Added a trait using the breast/cum system that makes Honey. I wasn't sure where honey would come from so it says breast.

Honey

**Changelog**
:cl:
- add: Added Honey producing trait.

### DIFF
--- a/Resources/Prototypes/_Floof/Traits/lewd.yml
+++ b/Resources/Prototypes/_Floof/Traits/lewd.yml
@@ -48,7 +48,7 @@
   - !type:AddCompsEffect
     components:
     - type: HoneyProducer
-      solutionname: "breasts"
+      solutionname: "honeysac"
 
 - type: trait
   id: SquirtProducer
--- a/Resources/Prototypes/_Floof/Traits/lewd.yml
+++ b/Resources/Prototypes/_Floof/Traits/lewd.yml
@@ -40,8 +40,6 @@
   description: trait-description-HoneyProducer
   category: Medical
   cost: 0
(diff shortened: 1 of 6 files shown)


Raikyr0:
> Looks okay, but I would recommend that this trait be incompatible with all the breast traits. Currently you can take normal breasts and honey with this build.

I was unsure where honey would come from, and defaulted to breast. I've left it un-exclusive if the individual using the trait IC expels honey from a different organ, like mouth or cock or whatever.

Resources/Prototypes/_Floof/Traits/lewd.yml
```diff
@@ -34,6 +34,24 @@
...
       solutionname: "breasts"
 
+- type: trait
+  id: HoneyProducer
+  name: trait-name-HoneyProducer
+  description: trait-description-HoneyProducer
+  category: Medical
+  cost: 0
+  conflicts:
+  - BigMilkProducer
```
SyaoranFox: Your definition excludes big milk, that is what I was referring to.

Raikyr0: Ah that I must have missed since I copied the milk trait.

Mnemotechnician (commented):
This is awful, you are duplicating a component, an event, and entity system logic for... no reason at all? This entire PR could be a total of 20 lines, instead it's 160 and introduces a new unnecessary update loop.

Just override MilkProducerComponent with a different produced reagent.

PR #319, opened 2026-03-27
Fishing - QoL

Raikyr0:
## About the PR
So I saw a complaint about fishing and people suffering physical real life pain from mashing. This PR makes it so you need to mash slightly less.

## Why / Balance
Real ass carp-al tunnel.

**Changelog**
:cl:
- tweak: Changed fishing to be easier on the wrist.

### DIFF
--- a/Resources/Prototypes/_Goobstation/Entities/Objects/Tools/fishing_rods.yml
+++ b/Resources/Prototypes/_Goobstation/Entities/Objects/Tools/fishing_rods.yml
@@ -26,7 +26,7 @@
       types:
         Piercing: 4
   - type: FishingRod
-    efficiency: 2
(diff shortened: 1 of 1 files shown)


Tarazicon:
carp-al tunnel
im sorry i really want this

Mnemotechnician (changes requested):
1. This is inconsistent. If you're increasing the effishiency of all fishing rods, multiply their efficiency by the same value, e.g. 1.5x.
2. You need to increase the UseDelay accordingly to keep the maximum fishing throughput the same.

Mnemotechnician (approved):
This is fine I suppose, since fishing efficiency currently only affects how much effort it takes to reel fish and not the actual fishing times

PR #321, opened 2026-03-27
Webbing/Rig Change

Raikyr0:
## About the PR
This PR, adds 50% explosion resistance to the security rig and mercenary webbing (just to keep in line) so tear gas can't get popped as easily. 
Partly reverts the Delta V webbing size change, so there's 16 blocks in 2 2x8 grids again up from 14 block in a 2x3 and 2x4 grid.
Removes the contraband tag from chest rigs and syndicate description to be more general (identical to old server).

Also changes the CMO Chest Rig to the old EE size.

## Why / Balance
1 - All the tear gas and stingers grenades popping at the very hint of explosion was making security not want to bring them. So this makes it very slightly safer.
2 - I don't know why Delta V split webbings into 2 blocks, and decreased the size of all of em by 2 blocks with the exception being the salvage rig. I think the split looks cool but the storage decrease wasn't needed, so the storage size is identical as it used to be.
3 - Changes the description to the old Floof description, and removed the contraband tag. Simply because it's not contraband, and the tag makes newer security have to re-read the contraband list.

Rig

**Changelog**
:cl:
- tweak: Added Explosion Resistance to Security Chest Rig and Mercenary Webbing.
- tweak: Changed Webbing and Rig storage space to be closer to what you're used to from the old server. Kept the block in the middle because it was cool.
- tweak: Changed Chest Rig description and removed contraband tag.


SyaoranFox:
I am gonna say no to this. Webbing should have a cost vs a belt. It is webbing, meaning it has gaps in it that can be pierced. Items at round start should be trade offs/sidegrades.

Raikyr0:
> I am gonna say no to this. Webbing should have a cost vs a belt. It is webbing, meaning it has gaps in it that can be pierced. Items at round start should be trade offs/sidegrades.

I'll work on this again. I was thinking something like having the webbing become research, and you needing to print it instead. (Using durathread and stuff.)

PR #330, opened 2026-03-27
Access Wire to Various Objects

Raikyr0:
## About the PR
This PR adds the Access Wire to
All Airlocks (Including High Security and Docks)
Anomaly Generator
Computers

## Why / Balance
Cutting the Access Wire makes the entity all access. For breaking and entering, currently one would disable the power to get in, thus leaving no doorlogs or emag the door or simply blow up the door or going through the wall. Doing this just gives the doorlogger more use. Engineering can also do this to set doors to all access in an emergency. It would return to what used to be normal on our old engine.

Adding the Access wire on the Anomaly Generator and Computers adds additional ways to sabotage or use something in an emergency without access.

accesswire

**Changelog**
:cl:
- add: Added Access Wire back to Airlocks, Computers etc

### DIFF
--- a/Resources/Prototypes/Wires/layouts.yml
+++ b/Resources/Prototypes/Wires/layouts.yml
@@ -10,6 +10,7 @@
   - !type:DoorTimingWireAction
   - !type:DoorSafetyWireAction
   - !type:AiInteractWireAction
+  - !type:AccessWireAction # Euphoria
 
 - type: wireLayout
   parent: Airlock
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Wires/layouts.yml
```diff
@@ -50,6 +51,7 @@
 - type: wireLayout
   id: HighSec
   wires:
+  - !type:AccessWireAction # Euphoria
```
Mnemotechnician: Adding an access wire to high-security airlocks is an odd choice. This kinda negates the fact that they are supposed to be high-sec.
```suggestion
```

Mnemotechnician (commented):
Other than that should be fine

Mnemotechnician (approved):
Now's fine.

PR #339, opened 2026-03-28
Spidercraft Arachne

Raikyr0:
## About the PR
Adds Spidercraft Tag to Arachne

I forgot if adding a tag overwrites the whole list from the parent so the whole list is there.

**Changelog**
:cl:
- add: Added Spidercrafting to Arachne


Mnemotechnician (changes requested):
Lacks comment markers

Mnemotechnician (approved):

PR #351, opened 2026-03-29
Shork

Raikyr0:
## About the PR
Ports the Shark Hard Toy, and finishes what I started by adding the other 2 sharks.
Added a funny contraband label for it. 

I'll do some testing, then consider where one would get this normally.

**Changelog**
:cl:
- add: Added Shork Hard Toy is back, with friends.

### DIFF
--- a/Resources/Locale/en-US/contraband/contraband-severity.ftl
+++ b/Resources/Locale/en-US/contraband/contraband-severity.ftl
@@ -54,3 +54,9 @@ contraband-examinable-verb-message = Check legality of this item.
 
 contraband-department-plural = {$department}
 contraband-job-plural = {MAKEPLURAL($job)}
+
+contraband-examine-text-Twink =
+    { $type ->
+        *[item] [color=#FFC5D3]This item is highly illegal bottom contraband![/color]
+        [reagent] [color=#FFC5D3]This reagent is highly illegal bottom contraband![/color]
+    }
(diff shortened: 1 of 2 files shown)


Raikyr0:
> So, I do have a concern with this. I have seen this spawn on station from maints, and from shonk spawn locations (such as on some of our beaches). Would you consider the damage excessive for something you can find randomly?

So, they weren't/aren't in any of the spawning pools, I can only assume a mapper added them in as a joke.

SyaoranFox (approved):

Mnemotechnician (changes requested):
Also these need a suffix like [DO NOT MAP] because we used to have one of those hard-mapped.

JFerrix (changes requested):
There is also the case of using an abstract class as basis, for example, each one of the toys uses the MeleeWeapon component with nearly identical values in most places, same with the DamageOtherOnHit component. A recommendation would be to expand the BaseTwinkContraband object to contain these components so you dont need to re-implement these for every toy, reducing code duplication and makes adding new ones easier for later. Not exactly world changing but I dont see a reason not to. Also where would these even show up? Because for random maint loot these, especially the blue one seems a bit busted. The pink one is fine by me honestly but the blue and gray one seem a bit on the strong end because pierce and slash are kinda strong. Even if they are comparable to forged swords, those require quite a few gameplay loops to go through. And if its an uplink item then having three nearly identical ones in there seems a bit redundant

SyaoranFox:
This has been open for a while, and has not received two approvals. As such, I am closing this PR. You are free to revise your idea, and re-implement this if you wish to.

PR #366, opened 2026-04-02
Adds Katana Belt to Security Loadout

Raikyr0:
## About the PR
Adds Katana Belt to security firearm loadout (because arming sword is in there).

I considered buffing the katana to deal extra limb damage, but a little out of scope at the moment.

The future plan would be to making it so katanas can dismember easier by adding a body part damage modifier to it, adding an odachi which should be 2 handed, which swings slower and has slightly longer reach. and has a back slot sheath.

## Why / Balance
Adds a more thematic weapon choice for Onis and some Kitsune, but anyone in general can take it.
And Vox have been eating too good.

**Changelog**
:cl:
- add: Added Katana Belt to Security Loadout (firearms)

### DIFF
--- a/Resources/Prototypes/_Floof/Loadouts/Groups/security_weapons.yml
+++ b/Resources/Prototypes/_Floof/Loadouts/Groups/security_weapons.yml
@@ -12,6 +12,7 @@
   - FloofWeaponLaserSvalinn
   - FloofPlasteelArmingSword
   - FloofWeaponEnergyGunMini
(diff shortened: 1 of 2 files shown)


M3739 (approved):
Thank you for your PR.

I had a long thought about this one, the daisho (sheathes), the katana and wakizashi itself, and utility aspects in comparison to the plasteel arming sword.

I think it's an interesting alternative to the lethal melee options we have available to our security officers. So, I'm for it.

PR #413, opened 2026-04-13
Reduce the scaling of point multipliers for epi.

SyaoranFox:
## About the PR

Title

## Why / Balance
Some Epi teams are pushing glimmer to its absolute limit to get points. This lowers the benefit of doing that by 1/5. 

## Media

Example Media Embed

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- tweak: Reduce the maximum glimmer multiplier from 2.5 to 2.

### DIFF
--- a/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
+++ b/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
@@ -66,7 +66,7 @@ public sealed class ArtifactAnalyzerSystem : SharedArtifactAnalyzerSystem
         if (sumResearch <= 0)
             return;
 
-        _glimmerSystem.Glimmer += sumGlimmer; // DeltaV - Add glimmer based on extracted points.    
+        _glimmerSystem.Glimmer += sumGlimmer; // DeltaV - Add glimmer based on extracted points.
         _research.ModifyServerPoints(server.Value, sumResearch, serverComponent);
         _audio.PlayPvs(ent.Comp.ExtractSound, artifact.Value);
         _popup.PopupEntity(Loc.GetString("analyzer-artifact-extract-popup"), artifact.Value, PopupType.Large);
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
This doesn't really deincentivize them, in fact it will only incentivize pushing further to achieve the maximum multiplier (since baseline artifact extractions give a puny amount of points)

SyaoranFox:
> This doesn't really deincentivize them, in fact it will only incentivize pushing further to achieve the maximum multiplier (since baseline artifact extractions give a puny amount of points)

Maybe. Currently they need to go to about 700-800 to get a decent return. However, with a reduction of 1/5, they would be dangerously close to triggering the 1000 glimmer event(s)

Mnemotechnician:
> Maybe. Currently they need to go to about 700-800 to get a decent return. However, with a reduction of 1/5, they would be dangerously close to triggering the 1000 glimmer event(s)

If you've hit 500 glimmer it's already too late. Unless you get a drainer, glimmer events like mites, revenants spawning, probers duplicateing, etc will drive it up.

JFerrix:
Ill be blunt with that, isnt that an entirely IC issue? Like, according to SLAW pushing glimmer to maximum or into ranges where glimmer events can happen can be interpreted as endangerment or the likes, which would (and should) give sec and/or the captain an obligation to reprimand or fire the mysta. And if the mysta is unaware they a) arent doing their job right and b) can and should reprimand or fire the responsible employees, I feel like modifying the game here will never change anything since thats a mindset and a relatively noteworthy RP/Law failure that I feel should be addressed rather.

Mnemotechnician:
Actually after further checking this won't even have any effect

image

SyaoranFox:
The problem that I was trying to solve with this was that epi research teams can be too fast with glimmer scaling, while also putting the station in a more dangerous position. Captains have been unwilling to curve Epi (for whatever reason), and technically Epi isn't behest to the station as they are NT. What I wanted to do was make that high glimmer state less appealing so that 1. IC it is a more dangerous thing to do, thus, causing more of a reaction, and 2. Players will determine that the risk isn't worth the reward.

Also, my understanding is that the cutoff should be at 2. Since its clamped to 2 rather than 2.5. I want to slow the progress of Epi so that it isn't a speedrun to raise glimmer, then explode 1 artefact.

PR #414, opened 2026-04-13
Lewd System Rework

Mnemotechnician:
## About the PR
Summary of the new systems:
- Each player can have a set of lewd organs. Each organ can produce, store, and deposit/receive fluids. Currently there are 4 types of organs. Rectum is the only one that's *new*, it can be used *only* for deposition (say gex or something idk).
- These organs can be picked and modified in the trait menu.
- In the right-click menu, inside the "lewd" category, the player can see various verbs related to lewd interactions. Those require both the user and the target to consent to it (non-players auto-consent). The exact verbs the user may see depends on the pair of (user organs, target organs), and currently include 3 major categories:
  - Milking (yourself or others, latter requires consent from target)
  - Sucking
  - The third kind of intercourse I don't know how to describe in SFW terms. 
  - I don't know what else to add, other players may figure it out.
- Organs that produce fluids only produce them when the user is neither starving nor parched, but do not currently consume either hunger nor thirst. This may be changed if players abuse it (which is unlikely because the actions for milking and the like purposefully have long do-afters so they are notorious to use outside of ERP scenes).
- Organs that produce liquids put the user's DNA into the data of the produced reagent. This can't be seen in-game yet, but it's the same thing that BloodstreamSystem does with blood, and you can see all the different reagents (but not their DNA) by collecting and putting them inside a chemmaster.
- A user can examine themselves to see what lewd organs they have and what are their fill percentages. This is hidden in a separate menu and visible only to the user.
- Organs that receive liquids (the bottom ones) slowly drain, leaking the received fluids onto the ground. This can be disabled with a trait.

Lewd interactions are visible to the target and user. They are also visible to others within a low 4m range, excluding ghosts, who can't see them at all (for their own sanity).

## Why / Balance
The main goals of this rework are to:
1. Replace the obsolete LewdTraitSystem with a more flexible and capable system.
2. Add some gameplay effects to ERP scenes and things alike.

## Technical details
LewdOrganSystem manages the addition/removal of lewd organs (relying on the standard OrganSystem). LewdSolutionsSystem manages the production and drainage done by lewd organs. LewdInteractionsSystem manages the addition of the relevant interaction verbs (relying on the standard InteractionVerbSystem).

LewdInteractionMapPrototype stores a dictionary, where keys are pairs (user organ, target organ), and values are lists of verbs. Whenever user right-clicks a target, the system checks what organs the user and target has, forms all possible combinations of them (including cases like (None, target organ) and (user organ, None)) and adds all verbs from all entries that match these pairs.

The actual verbs are defined using the InteractionVerbSystem. So are their requirements and actions.

Summary of changes to misc systems:

  InteractionVerbSystem:

  - Verbs can now automatically repeat if proto.Repeat == true and args.AllowRepeat == true (default, actions or callers can override it)
  - Verb effects and do-after args now have [AlwaysPushInhertance] so they don't have to be re-defined if you just want to change one field
  - Added a new VerbCategory prototype, which can be used to dictate which category interaction verbs are placed into.
  - InteractionPopupPrototype now has an AllowGhosts field, which can be used to obscure certain popups from ghosts. Lewd popups use that.

  TraitSystem:

  - Traits now have a UsesSlots property, setting it to false does exactly what you'd expect.

  ModifyUndiesSystem:

  - Exposed an API method to check if an entity has hidden underwear, this one is pure shitcode.

## Media

TODO

Old demonstration:

https://github.com/user-attachments/assets/fc5db7b8-9467-4c05-80d9-7a9da4276186

New demonstration:

I couldn't find a way to compress the video enough, so here's a 2x sped up version of the full demo: https://drive.google.com/file/d/187nBIH52HHa6734CXn5EFFCuhy9zMtJO/view?usp=sharing

</p></details>

## Licensing:
    

## Breaking changes
See technical details

**Changelog**
:cl:
- add: Reworked lewd traits. There's now a set of 4 base traits that give you special organs that can produce or store liquids, and several traits that modify their stats. Players who have chosen one of the existing lewd traits will have to choose again.
- add: There's 12 new lewd interaction verbs available that are related to the lewd traits. In order to use them or have others use them on you, you need to enable a couple new consent toggles. Most of them require you to have no inner clothing and no matching underwear (use the right click menu to remove underwear markings if you have any). Keep in mind that these interactions can be seen by others (within a 4m radius).
- add: Certain traits can now be excluded from the global trait limit. Lewd traits are an example of that.
- fix: The "help up" interaction should now work correctly.


Mnemotechnician:
It is... mostly done. I tried to come up with reasonable popup texts for these interactions, but if anyone wants to suggest their own, it's best to do now: [Resources/Locale/en-US/_Floof/interaction/verbs/lewd/lewd-interactions.ftl](https://github.com/Mnemotechnician/Panta-Rhei/blob/feat/better-lewd-system/Resources/Locale/en-US/_Floof/interaction/verbs/lewd/lewd-interactions.ftl)

Mnemotechnician:
I'll do the last round of testing and merge this if no issues arise, since there are future prs depending on this.

Mnemotechnician:
It is time. Test fail is unrelated.

PR #425, opened 2026-04-15
Minor trait tweaks

UtriculariaCalycifida:
## About the PR

Chronic pain popups are only for the person with the trait, and numbness now gives 2 points.
## Why / Balance

Not being able to estimate how much health you are at kinda sucks, so -2 seems good to me. I was losing my mind from seeing the popups.
## Technical details

YAML.
## Media
No media for the popups. I can't really show something not being there.

image
</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- tweak: Numb now gives 2 points rather than costing one.
- tweak: Only the user can see chronic pain popups.


Mnemotechnician (approved):

Mnemotechnician (approved):
Ideally the SharedChronicPainSystem shouldn't be running on the client side at all, and ChronicPain should have SendOnlyToOwner = true, but idc enough to fix it.

PR #437, opened 2026-04-17
Alternative Medical PDAs

Dunrab:
## About the PR
Some more alternative medical PDAs to give more flavor to some of the roles.

## Why / Balance
I like medical and want more pdas, besides whos gonna stop me?

## Technical details
The usual sprites, yaml and ftl edits.

## Media

Example Media Embed

</p></details>

## Requirements

## Licensing:

**Changelog**

:cl:
- add: Senior Physician, Medical Doctor and Parameds have some additional alternative PDAs
- fix: Senior Physician no longer has access to the medical doctor PDAs.


SyaoranFox:
I do not think its a wise idea to use the keyword "senior" as part of any title that isn't a senior role. Adding this here that circumvents the timer means that someone could be labelled as "experienced" when in fact they are not.

Dunrab:
> I do not think its a wise idea to use the keyword "senior" as part of any title that isn't a senior role. Adding this here that circumvents the timer means that someone could be labelled as "experienced" when in fact they are not.

Uh the senior phys pdas are for the senior role, i had to make a second one because the original is called mentor and has a 50 hour time lock behind it

SyaoranFox:
> > I do not think its a wise idea to use the keyword "senior" as part of any title that isn't a senior role. Adding this here that circumvents the timer means that someone could be labelled as "experienced" when in fact they are not.
> 
> Uh the senior phys pdas are for the senior role, i had to make a second one because the original is called mentor and has a 50 hour time lock behind it

Fair. Rather than duplicate the PDA, we should reduce the time associated with the mentor role. Or, we could change the mentor role to something like "Medical Consultant"

Dunrab:
> > > I do not think its a wise idea to use the keyword "senior" as part of any title that isn't a senior role. Adding this here that circumvents the timer means that someone could be labelled as "experienced" when in fact they are not.
> > 
> > 
> > Uh the senior phys pdas are for the senior role, i had to make a second one because the original is called mentor and has a 50 hour time lock behind it
> 
> Fair. Rather than duplicate the PDA, we should reduce the time associated with the mentor role. Or, we could change the mentor role to something like "Medical Consultant"



> > > I do not think its a wise idea to use the keyword "senior" as part of any title that isn't a senior role. Adding this here that circumvents the timer means that someone could be labelled as "experienced" when in fact they are not.
> > 
> > 
> > Uh the senior phys pdas are for the senior role, i had to make a second one because the original is called mentor and has a 50 hour time lock behind it
> 
> Fair. Rather than duplicate the PDA, we should reduce the time associated with the mentor role. Or, we could change the mentor role to something like "Medical Consultant"

Yeah removed the duplicate and just lowered delta-vs time requirement so when a player unlocks senior physician they also unlock the pda to be used as a med doc and that senior role

Mnemotechnician (approved):
Tis crazy

PR #441, opened 2026-04-18
Added Potential Psionic to Silicon Base, and Offer Item to IPC and Arachne base.

Raikyr0:
## About the PR
So someone was wondering why the Oracle wasn't talking to their Psionic IPC character, and as it turns out, even if the trait that gives an IPC psionics makes them a Psionic, it doesn't give them PotentialPsionic which allows the oracle to talk to them.

This change allows them to talk to the Oracle. Also technically, they were immune to the effects of tinfoil hats and most likely could cast wearing them.

This PR also fixes IPCs and Arachne missing the Offeritem Comp.

**Changelog**
:cl:
- fix: Fixed IPCs missing a component that allowed them to talk to the Oracle, despite being psionic.
- fix: Fixed IPCs and Arachne not being able to offer things.

### DIFF
--- a/Resources/Prototypes/_EE/Entities/Mobs/Player/ipc.yml
+++ b/Resources/Prototypes/_EE/Entities/Mobs/Player/ipc.yml
@@ -115,6 +115,8 @@
     canDisarm: true
   - type: SnoutHelmet # DeltaV - IPC Snouts
     replacementRace: vulpkanin
+  - type: Hands # Floofstation
+  - type: OfferItem # Floofstation - Fixes Offer for IPCs
 
 - type: entity
   save: false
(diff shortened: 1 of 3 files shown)


Resources/Prototypes/_EE/Entities/Mobs/Player/ipc.yml
```diff
@@ -115,6 +115,8 @@
     canDisarm: true
   - type: SnoutHelmet # DeltaV - IPC Snouts
     replacementRace: vulpkanin
+  - type: Hands # Floofstation
+  - type: OfferItem # Floofstation - Fixes Offer for IPCs
```
Mnemotechnician: They already have both

Raikyr0: They don't. They use PlayerSiliconHumanoidBase which doesn't parent off BaseMobSpecies.
A search shows that ONLY BaseMobSpecies has OfferItem.

Mnemotechnician (approved):

PR #457, opened 2026-04-23
Barvend fix

Raikyr0:
## About the PR
I don't know why this was a thing but DV's command and unlocked boozomat had DIFFERENT Inventories. So it was missing a ton of things, most memorable, cum.

**Changelog**
:cl:
- fix: Fixed the Unlocked/Command Boozomat having a DIFFERENT Inventory.

### DIFF
--- a/Resources/Prototypes/_DV/Catalog/VendingMachines/Inventories/commandboozeomat.yml
+++ b/Resources/Prototypes/_DV/Catalog/VendingMachines/Inventories/commandboozeomat.yml
@@ -1,39 +1,60 @@
 - type: vendingMachineInventory
   id: BoozeOMatCommandInventory
   startingInventory:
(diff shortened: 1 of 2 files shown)


SyaoranFox (approved):

Mnemotechnician:
The command bar having a limited stock is intentional. They aren't supposed to have their own bar, just a small subset of it.

Raikyr0:
> The command bar having a limited stock is intentional. They aren't supposed to have their own bar, just a small subset of it.

I don't think it gets used that often to start, as command tend to just use the normal bar.

SyaoranFox:
Due to not having a second approval, and no one else has found this to be an issue, I am closing this PR. If this should prove to be false, please resubmit your changes for review.

PR #459, opened 2026-04-23
Add Tiny Trait

kivdon:
## About the PR

Adds the Tiny trait, like we had on Floof. It allows you to fit in 2x2 storage spaces, but also comes with debuffs.

Note: The trait reduces the number of free hands required to be picked up to 1. This works fine if the picker-upper has an item in one hand, and a free hand (the Tiny person getting picked up takes up that free hand, the other hand remains holding the item). But if the picker-upper has both hands free, the carried person takes up both hands. This doesn't break anything, and probably will never be noticed since it only affects those with the Tiny trait. I just wish I could get it so a Tiny person only takes up one hand, even if both hands are free. I may work on that later.

## Why / Balance

This has been a long requested trait, similar to Small. The Discord threads for Small and Tiny cover this.
Balance:
I've reduced Small to a cost of -1 (was -2, the advantage of being able to fit in only things of duffel size or above is not enough for a -2), and set Tiny to be -2. 
On Floof, Tiny was a cost of 0, but had strict requirements for being able to use it, and it had heavy debuffs. 
Here, I've implemented some of the debuffs, such as dealing less melee damage, and having less stamina. But since I didn't put some of the other debuffs on it, I've given it a point cost. If anything, I worry that the debuffs might be too much for its point cost. 
I did not put size requirements on the trait, for RP reasons. 

## Technical details

Just trait yml changes. 

## Media

image

image

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- add: Added Tiny trait! Go climb into backpacks and satchels, but just beware that you're not as strong as you were.

### DIFF
--- a/Content.Shared/_DV/Carrying/CarryingSystem.cs
+++ b/Content.Shared/_DV/Carrying/CarryingSystem.cs
@@ -304,7 +304,7 @@ public sealed class CarryingSystem : EntitySystem
         // Floofstation - store virtual items and add a special component to them
         if (_virtualItem.TrySpawnVirtualItemInHand(carried, carrier, out var virt))
             EnsureComp<OfferableVirtualItemComponent>(virt.Value);
-        if (_virtualItem.TrySpawnVirtualItemInHand(carried, carrier, out virt))
+        if (TryComp<CarriableComponent>(carried, out var carriableComp) && (carriableComp.FreeHandsRequired > 1) && _virtualItem.TrySpawnVirtualItemInHand(carried, carrier, out virt))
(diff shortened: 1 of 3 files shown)


Forzii:
If you look at my original implementation (I know there were multiple PRs, but the last one should be a full roll-up of everything cause I had to re-add it after it got pulled briefly) I'm pretty sure I fixed that free hands required bug. Feel free to poke me if you have any questions about the original tiny trait.

Edit: Oh, it looks like Mnemo made the height adjust system ya'll are using now, that probably means it's a lot less broken and finicky than the old EE one.

PR #460, opened 2026-04-23
Biomass Reclaimer Fix

Dunrab:
## About the PR
Sometimes the biomass reclaimer would not allow the player to biomass a mob. At first i was using pulse weapons because haha funny mob dies quick but every time id kill a mob like this the reclaimer would work with out issue. So then I decided to swap to weapons security would normally use in a round and tabbed out to respond to someone in discord, during that brief tab out time the mobs blood volume reached zero, i attempted to biomass and it failed like we were seeing in rounds. I then gave it some blood and it let me biomass it again.

## Why / Balance
Med needs to be able to biomass mobs.

## Technical details
This hurt my head a lot and i spent more time than i should have on this. Basically when a mob had no blood, component.BloodReagents.Volume would equal zero. component.BloodReagents.ScaleSolution(50 / component.BloodReagents.Volume); would then try to divide by zero, thus preventing the mob from being biomassed. To fix this i just changed component.BloodReagents.Volume to component.BloodReagents.MaxVolume and when the mob has no blood it can still be biomased and arent crashing in debug from dividing by zero anymore. This is probably a really bad solution the problem but its what i came up with so here we are.

## Media

The discovery:
https://github.com/user-attachments/assets/3626ac9c-9926-4417-b23b-357d14e110a1

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- fix: The biomass reclaimer should now be able to biomass mobs again.

### DIFF
--- a/Content.Server/Medical/BiomassReclaimer/BiomassReclaimerSystem.cs
+++ b/Content.Server/Medical/BiomassReclaimer/BiomassReclaimerSystem.cs
@@ -212,12 +212,12 @@ namespace Content.Server.Medical.BiomassReclaimer
             {
                 component.BloodReagents = solution.Clone();
                 /* Floofstation change:
-                 * was component.BloodReagents.Volume, changed to component.BloodReagents.AvailableVolume
-                 * Was changed due to if a mob has 0% blood, ccomponent.BloodReagents.Volume would be 0 this would lead to us dividing by zero and the mob would not be biomassable.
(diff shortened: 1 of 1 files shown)


Mnemotechnician (approved):

Mnemotechnician:
Fun fact, if i understand the formula right, the reclaimer used to spit out blood inversely proportional to how much the mob had

M3739 (approved):

PR #462, opened 2026-04-24
Add Measurements to the Height Slider

Cprn-Sprkl:
## About the PR

Adds measurements to the height slider in CM and feet/inches. Measurement conversions are from a chart shared in the community.

This does not tie in to the records screen on purpose. This allows a little extra wiggle room at the max and min values for people to input their own canon height. Say, someone plays a 242cm character. That's out of the slider range, but they can just make their official record show that height.

## Why / Balance

Standardizes the height slider so everyone is on the same page on what 0.81 means compared to 0.95. Minimizes situations where someone who plays a five foot character towers over someone playing a five foot two character.

## Technical details

Adds a label to the UI that displays the measurements. Adds a code block to do some math processing. It is a little more complicated than it should be, but apparently Robust implemented its own toString formatting that doesn't recognize "D2" or ":D2" so I did a work around with my more math inclined partner.

## Media

image
image
image

</p></details>

## Requirements

## Licensing:

## Breaking changes

N/A

**Changelog**

:cl: sprkl
- add: Height measurements now display by the height slider.

### DIFF
--- a/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml
+++ b/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml
@@ -80,6 +80,7 @@
                                     <Control HorizontalExpand="True" />
                                     <Slider Name="CDHeightSlider" HorizontalAlignment="Right" SetWidth="300" MinValue="0.0" MaxValue="1.0"/>
                                     <LineEdit HorizontalAlignment="Right" Name="CDHeight" MinSize="60 0" Text="1.0" />
+                                    <Label Name="EuphHeightDisplay" HorizontalAlignment="Right" MinSize="60 0"/>
                                     <Button Name="CDHeightReset" Text="{Loc 'humanoid-profile-editor-reset-height-button'}" HorizontalAlignment="Right"/>
                                 </BoxContainer>
                                 <!-- End CD - Character Records -->
--- a/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
+++ b/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
@@ -44,7 +44,9 @@ using Content.Shared._CD.Records;
 using Content.Shared.Chemistry.Reagent;
 using Content.Shared.FixedPoint;
 // End CD - Character Records
-using Content.Shared._DV.Traits; // DV - Traits
+using Content.Shared._DV.Traits;
+using System.Runtime.InteropServices;
+using Robust.Shared.Toolshed.Commands.Values; // DV - Traits
 
 namespace Content.Client.Lobby.UI
 {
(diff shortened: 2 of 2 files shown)


Arbitrated:
Personally, we don't like this implementation, because we've always felt like that chart in particular is very... Weird. The sprite scaling features are based on a default size, so for example a 0.9x size person would be 0.9x the size of a "normal" person - the chart instead reduces their height by 10 inches (result: 4'11" or 149.86 cm). 

In slightly more mathematical terms, 5'9" (69 inches, or 175.26 cm) (the default) is 1.0x, and 0.9x scale should result in a character being 62.1 inches, 157.734 cm, or 5'2".

This has always bothered us, and will continue to bother us, on all the servers that's referenced the chart. The IDEA behind this PR is good, and we love the idea, but the math bothers us _**a lot**_.

Mnemotechnician:
Average sized human should be 171 cm, but according to your screenshot it's 130/0.82 = 158.5 cm, which is 10% shorter. Also, you're not accounting for the fact that different species may have different heights - EE worked around it by introducing an "average height" to the species prototype

Cprn-Sprkl:
Average size for human is 1.0, which comes out to 175cm or 5'9".

Species do have different sizes. When you select and save a species and then hit reset on the height slider it puts it at the default for the species. Most use 1.0. Harpy, felinid, and rodentia use 0.8. Reptilians use 0.95. Oni use 1.2. That's how it works by default. Here is several species lined up and spawned in a straight line, only moving right:
image
That's Diona, Reptilian, Human, Oni, Diona (again due to misinput), Rodentia, Moth, and Harpy. Each at the default size for their species.

This is a Human set at 1.2 and an Oni set at 1.2:
image
They are identical in size. So if you make your human be 1.2 it will be 226cm or 7'05". If you make your Oni 1.2 it will be 226cm or 7'05". This is how the height slider is already working here and not a change introduced in this PR.

Here is a human set at 0.8, it is identical to the default rodentia and harpy, which are 0.8:
image

Mnemotechnician:
> Average size for human is 1.0, which comes out to 175cm or 5'9".

In your screenshot, a 0.75x human is told to be 112cm, and a 0.82x one is told to be 130.
```
112 / 0.75 = 149.333    (the correct value should be 175, and the height of a 0.75x human should be 131.2)
130 / 0.82 = 158.537    (ditto, and the height of a 0.82x human should be 143.5)
```
Something is awfully wrong in your math - you're achieving an error of almost 15% at 0.75x scale, likely because you're using inches in the code instead of cm. Also, 5'

> Species do have different sizes. When you select and save a species and then hit reset on the height slider it puts it at the default for the species. Most use 1.0. Harpy, felinid, and rodentia use 0.8.

That's not the point. Species can be inherently different in visual size (and physical size, but that's not something that can be easily compared or judged in-game). For example, a resomi at 1x scale is naturally shorter (has a smaller sprite) than a human at the same scale. Dwarves used to be shorter than humans, but they were disabled in favor of configurable height afaik.

Cprn-Sprkl:
I'm guessing from your math the PR description was simply skimmed.

image
The math does what it says in the PR description. Maps the height slider to this chart to give some kind reference for players.

But you're right. Resomi are the only currently implemented species with a different height sprite to humans and so they break this. That's that.

PR #473, opened 2026-04-26
Ninja exit condition

Raikyr0:
## About the PR
Adds a Ninja Escape Implant (Death Acidifier) Implanter exists too but obviously no normal way to get this.
It deploys a smoke cloud and leaves a Ninja Mask (Will change this to a sassy Spider Clan calling card style note if someone ever makes it).

## Why / Balance
Hostile Ninjas typically, do their objectives, greentext and then have no exit condition other then play "ring around the rosie" with a often frustrated security, which leads to a lot of OOC complaining.

This PR adds a "exit condition" which is, leaving the station in a poof of smoke, mission complete, jobs done instead of simply dying or getting captured on purpose. 

## Media

https://github.com/user-attachments/assets/d5aa0826-6fbb-4d18-bf74-54450a2214ae

## Requirements

**Changelog**
:cl:
- add: Added an exit conditions for Ninjas so they don't feel the need to linger around on station.

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -291,7 +291,7 @@
         - Syndicate
       - type: AutoImplant
         implants:
-        - DeathAcidifierImplant
+        - NinjaEscapeImplant
       - type: RandomMetadata
         nameSegments:
         - NamesNinjaTitle
--- a/Resources/Prototypes/_Euphoria/Actions/types.yml
+++ b/Resources/Prototypes/_Euphoria/Actions/types.yml
@@ -1,0 +1,16 @@
+- type: entity
+  parent: BaseSuicideAction
+  id: ActionActivateNinjaEscape
+  name: Activate Ninja Escape
+  description: Activates your Ninja Escape implant, teleporting you far-far away from this station. You won't be able to come back. Use this when your job is done and you need to return to base.
+  components:
+  - type: Action
+    checkCanInteract: false
+    checkConsciousness: false
+    itemIconStyle: BigAction
+    priority: -20
+    icon:
+      sprite: _Goobstation/Textures/Effects/bluespace_lifeline.rsi
+      state: bluespace_lifeline
+  - type: InstantAction
+    event: !type:ActivateImplantEvent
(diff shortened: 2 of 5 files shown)


Mnemotechnician:
This is cool but I feel like it should be a separate implant, not a replacement. Acidifying is the last resort option in case the ninja gets captured, this should work only if the ninja is free (see ActionComponent.checkCanInteract and checkConsciousness)

M3739:
> This is cool but I feel like it should be a separate implant, not a replacement. Acidifying is the last resort option in case the ninja gets captured, this should work only if the ninja is free (see ActionComponent.checkCanInteract and checkConsciousness)

I second this, a separate implant would be best. Ideally, I would love for this implant to, if possible, only become available when a greentext is indeed achieved. Because I know someone is going to prefer this implant over the normal acidifier due to the fact it does leave behind only the mask.

Raikyr0:
> This is cool but I feel like it should be a separate implant, not a replacement. Acidifying is the last resort option in case the ninja gets captured, this should work only if the ninja is free (see ActionComponent.checkCanInteract and checkConsciousness)

Roleplay wise - So, in my head at least, I feel like a Ninja could just slip out of cuffs (because a Ninja would totally be trained to do that) to poof escape, but for gameplay purposes, giving em a freedom implant would give security a stroke.

Coding wise - Issue would be, they both count as acidifiers, I already had to do some jank to get a delay working so the smoke goes off before they "gib". If a normal acidifier was added into the equation, I think the Ninja one wouldn't pop at all.

SyaoranFox:
> > This is cool but I feel like it should be a separate implant, not a replacement. Acidifying is the last resort option in case the ninja gets captured, this should work only if the ninja is free (see ActionComponent.checkCanInteract and checkConsciousness)
> 
> Roleplay wise - So, in my head at least, I feel like a Ninja could just slip out of cuffs (because a Ninja would totally be trained to do that) to poof escape, but for gameplay purposes, giving em a freedom implant would give security a stroke.
> 
> Coding wise - Issue would be, they both count as acidifiers, I already had to do some jank to get a delay working so the smoke goes off before they "gib". If a normal acidifier was added into the equation, I think the Ninja one wouldn't pop at all.

It may be that you need to make a new branching implant that does what you want rather than piggybacking the acidifier. Your idea here is really solid, and I think it deserves to be fully developed.

SyaoranFox:
This PR requires more thought and work for it to be viable. The idea is sound, and the concept makes sense. If you wish to take our feedback onboard and have another attempt, then feel free to PR this again.

PR #474, opened 2026-04-26
Elite suit unshittening

Raikyr0:
## About the PR
Reverts DV's "Side-grade" to the Elite suit and slightly improves it in a non-major way.

## Why / Balance
So, just to explain, DV says it's a "Side-grade", it was a downgrade, in both doubling the cost and armor.
This reverts the change so it's a "Side-upgrade" so it's the Atmos Suit of Syndies.
Reduced the price back to EE's 10 down from 14... which was insane.
Also minor DV related buffs is just improvements to cold damage.

## Requirements

**Changelog**
:cl:
- tweak: Made the Elite/Thermal Suit good again, it was too expensive for "a side DOWN-grade".

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
@@ -842,8 +842,8 @@
   - type: Armor
     modifiers:
       coefficients:
-        Blunt: 0.6 # Euphoria - was 0.2
-        Slash: 0.6 # Euphoria - was 0.2
+        Blunt: 0.6 # Euphoria - was 0.8
+        Slash: 0.6 # Euphoria - was 0.8
         Piercing: 0.6
         Heat: 0.2
         Radiation: 0.01
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
@@ -848,7 +848,7 @@
         Heat: 0.2
         Radiation: 0.01
         Caustic: 0.5
-        Cold: 0.2 # Euphoria - was 0.3 // In the unlikely event that syndies end up fighting coscult
+        Cold: 0.2 # Euphoria - was 0.4 // In the unlikely event that syndies end up fighting coscult
   - type: DamageOnShootProtection # imp
     slots: OUTERCLOTHING
     damageProtection:
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
@@ -826,7 +826,7 @@
       outerClothing-dog:
       - state: equipped-OUTERCLOTHING-dog
   - type: PressureProtection
-    highPressureMultiplier: 0.02
+    highPressureMultiplier: 0.001 # Euphoria - was 0.02
     lowPressureMultiplier: 1000
   - type: TemperatureProtection
     heatingCoefficient: 0.001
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
```diff
@@ -826,37 +826,39 @@
       outerClothing-dog:
       - state: equipped-OUTERCLOTHING-dog
   - type: PressureProtection
-    highPressureMultiplier: 0.02
+    highPressureMultiplier: 0.001 # Euphoria - was 0.02
```
Mnemotechnician, line 829: Why does a nukie to be able to withstand 4000 atmospheres (4 GIGAPASCALS) of pressure?

Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
```diff
@@ -826,37 +826,39 @@
...
+    damageCoefficient: 0.5 # Euphoria - was 0.3
+  - type: ZombificationResistance
+    zombificationResistanceCoefficient: 0.1 # Euphoria - Added 90% Zombie Resistance // This effect will only ever show up in planned events on Euphoria
   - type: Armor
     modifiers:
       coefficients:
-        Blunt: 0.8 # DeltaV - was 0.6
-        Slash: 0.8 # DeltaV - was 0.6
+        Blunt: 0.6 # Euphoria - was 0.2
+        Slash: 0.6 # Euphoria - was 0.2
```
Mnemotechnician: Was 0.8

Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
Raikyr0: So, it's unlikely to ever reach a pressure point that damages Atmos and Elite suits normally.
Even in a burn chamber, considering they tend to self regulate and stop before it reaches a point where the pressure can actually damage it in regular gameplay. (Outside of the old EE TEG burn which had it go up to millions of KPA, but I don't believe it's possible anymore.)

Mostly a superficial change to assure it's truly pressure immune.

Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
```diff
@@ -826,37 +826,39 @@
...
-        Blunt: 0.8 # DeltaV - was 0.6
-        Slash: 0.8 # DeltaV - was 0.6
+        Blunt: 0.6 # Euphoria - was 0.8
+        Slash: 0.6 # Euphoria - was 0.8
         Piercing: 0.6
         Heat: 0.2
         Radiation: 0.01
         Caustic: 0.5
-        Cold: 0.4 # DeltaV - Cold res
+        Cold: 0.2 # Euphoria - was 0.3 // In the unlikely event that syndies end up fighting coscult
```
SyaoranFox: Was 0.4

SyaoranFox (changes requested):
I really don't understand your changes to make the suit unshittening. You made the suit less resistant to blunt, slash, and cold while removing its speed penalty. You give them more heat resistance and over do the pressure multiplier. I actually think this isn't achieving what you want it to.

Raikyr0:
> I really don't understand your changes to make the suit unshittening. You made the suit less resistant to blunt, slash, and cold while removing its speed penalty. You give them more heat resistance and over do the pressure multiplier. I actually think this isn't achieving what you want it to.

From my understanding at least 0.6 means its 40% resistant, which is up from 0.8 - 20%

Juggsuit has 0.3 Piercing, which means it has 70%.

SyaoranFox (approved):

SyaoranFox:
I just tested this. I approve for now, but if it starts becoming a problem we will look at reversing it. Needs one more approval.

JFerrix (approved):
Lets give it a shot

PR #475, opened 2026-04-26
Content Warning: Vore patch 1

JaDiDo:
## About the PR

Making first adjustment to the code such as dealing with bags/holding, preventing exploits and reordering some values to make more sense

## Why / Balance

Making sure that no one uses the system to get immunity such as spacebreathing, dealing with more scenarios that could happen (infamous holding bug from the system before rebase that made you visible) and code readability 

## Technical details

1. There is now a check if the target is inside the designated container inside the immunities method which should prevent exploitation to get those without actually being inside
2. Targets will now be ejected of containers (like bags) that arent the designated container inside the doAfter method before being moved to the designated one (check EnsureEntityfree)
3. Target that is either (1.) being held by user, (2.) having the user being held or (3.) is being hold by a third party member will now be dropped during the attempt (check EnsureEntityfree)
4. MaxPrey will now only count folks with bodycomponents and not items. Also altered the value and if check to make more sense (previous version said 2 but you can have up to 3, now it counts correctly)
5. Entities will now be checked on the initialization if they had a bodycomponent to avoid giving items vorecomponent altogether (was previously in ongetverbs)
6. being devoured popup is now marked as a largecaution (been told and feel myself that the small one gets neglected too easily)

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- tweak: adjusted the voresystem


SyaoranFox (approved):
We are now free from matryoshka ops.

M3739 (approved):
LGTM.

PR #490, opened 2026-04-29
fixes cost on synth trait

Raikyr0:
## About the PR
Fixes the cost on the synth trait to cost 0, as its additions are flavor.

**Changelog**
:cl:
- fix: Fixed Synth Trait costing a point when it should have been neutral.

### DIFF
--- a/Resources/Prototypes/_DV/Traits/medical.yml
+++ b/Resources/Prototypes/_DV/Traits/medical.yml
@@ -97,6 +97,7 @@
   name: trait-synth-name
   description: trait-synth-desc
   category: Cybernetics # Euphoria
+  cost: 0 # Euphoria - Missing Cost meant it costed 1
   conditions:
   - !type:HasCompCondition
     component: Silicon


M3739 (commented):
I have a burning question.

In your PR, you state the following:

> Fixes the cost on the synth trait to cost 0, as **<ins>its additions are flavor.</ins>**

 However, on the tin of this trait, it states the following:

> You are a biomechanical construct, who bleeds coolant and is notified of **<ins>ongoing Ion Storms</ins>**.

Because of this description, to the player, this trait is _not purely flavor_. The ability to sense an ion storm is beneficial in helping the crew ensure silicon laws are intact. It enables the detection, with reasonable certainty, of altered laws without having to rely on intuition, analysis, or investigation.

Ergo, this trait is *not* just flavor. It serves a _functional purpose_. Just because the rate of detecting an Ion Storm with this trait is only 30% does not diminish the fact it is functional to begin with. When the 30% does hit, you gain _actionable intelligence._ That can alter the course of the round.

So, why, then, would we make this trait free?

Arbitrated:
Wasn't this already a PR at some point? What happened to that one?


Anyways, this trait was originally designed for a HRP server, which doesn't have any trait "builds" or "points" in the first place, and this concern was never needed there. The Ion Storm notification feature was intended for allowing characters with the Synthetic trait to hamper themselves (with or without admin help), under the logic that, without "Laws", an ion storm could still do some damage/interfere with synthetic people. We've seen cases of organ failure, dizziness, ignitions, and more.

As we mentioned when this trait getting fixed was PREVIOUSLY brought up, if a character is going to use the ion storm notification for harassing Borg/AI players instead of personal roleplay, then that is a major player issue that should be ahelped or otherwise reported. 

As a comparison, the recent "lewd traits" added to Euphoria also provide a functional, non-flavor purpose, if producing reagents - milk, lubricant, and semen, which are useful especially in the service department. Most of these traits are also Zero cost (with the most "intense" versions of these traits being still very cheap, but costing points), but the reagents **can** be used to alter the course of the round.

M3739:
Fair point. I acknowledge that the reasoning I gave does indeed make the recent lewd updates fall under that same umbrella.

Allow me to elaborate further.

If someone, was to say, use the lewd system as a way to run recipes to alter the round in some manner, or in a manner which is, but not limited to, deemed either a consent violation or to forward antag objectives, Admins will undoubtably sanction the offending player. There is no long-term pattern detection or intuition required other than confirming the fact the violation happened in the first place. Sure, it is indeed functional with flavor under my blanket description, but it's mitigated by the fact that misuse often comes with swift and severe disciplinary action, and only requires a single incident. As a result, players often do not misuse such systems because all it takes is a single violation to result in sanctions. Misusing it can clearly, easily, and readily be interpreted as _illegitimate_ usage.

If someone was to use the ion storm detection capability of the synth trait to always act on the positive detection of a Ion storm, the issue is the fact that such actions, to the eyes of the average player and likely admins, is _legitimate_. You can say your just doing your job as an MG, keeping your borgs in check as a Roboticist, or, doing what your job description says as a Security Technician.

At that point, it then begs the question; _'Is it really harassment if it's just someone doing their job?'_ Sure, you can say that, if the treatment is not across all silicons, and the discrimination is not *just* or *due*, then yes, I would think that is harassment. Coming to that conclusion is a different story. One that requires keen observation, intuition, and pattern-recognition, which I can tell you not every admin has. As a result, it is not feasible to determine consistently and in a timely manner, whether or not someone is abusing the trait. Because unlike the lewd system, the line between legitimate and illegitimate use is very fuzzy at best as a result.

And the last thing I want is a key appeal factor of silicons, the prospect of ion laws and the intricacies there of, to be dimmed as a result of a singular trait. 

**The only implementation, where I can accept it being free, is if the ability to sense ion storms is cut entirely.** Reason being, is our unique player psychology, which includes admins, and that it doesn't make sense (here, at least) for a non-silicon-lawed synthetic or individual to be able to detect ion storms *directly*. Silicons do 'detect' them in normal gameplay by having their laws suddenly changed without player interaction.

Raikyr0:
Personally speaking, I've never seen a synth speak up about being ion stormed, and technically Thaven are also supposed to be capable of telling if an ion storm hit. For the most part, I think most people running Synth use it to roleplay as an advanced IPC with no battery.

M3739:
Same here. I have yet to witness a Synth roleplay stemming from a detected ion storm, and if there is an instance, I'd love to know about it.

Thaven, as far as I know, don't react to ion storms, I believe they react more to noospheric discharges or bursts. They got their own thing going on.

And I agree with the most part. Either advanced IPC, an android, or what I like to do, a "full-body prothesis". It enables for a flavor of synthetic that is more akin to emulating the living than IPCs being automatons from the ground up.

Raikyr0:
Swiftly closing this because I saw an OLDER PR doing the exact same thing go bother that one.

PR #492, opened 2026-04-29
Trait balance fixes

Raikyr0:
## About the PR
Adjusts/Balances Various Trait cost
Fixes Sechud/Omnihud
Increases the Cost of Pryarms and Regeneration which have been an issue partly

## Why / Balance
The LIST

- Amputee - -2 - Missing arm prevents wielding of weapon, even with small arms it prevents carrying a shield in the off-hand.
- Narcolepsy - -2 - Narcolepsy has no warning unlike the previous codebase.
- Redshirt - -2 - Redshirt removes crit state, and in certain scenarios makes your revival impossible.
- Precognition - -1 - Precog is niche and somewhat useful however, it's use increases the glimmer and pisses Epi off.
- Psychokinetic Scream - -3 - Psychokinetic Scream is very niche, annoys janitors and gas lights people into thinking there's a Skia.
- Psionic Regeneration - +2 - Psionic Regeneration has had much discussion, it's downside being "slowed to a crawl" and requires some skill to use wasn't enough, as people simply just hunkered down, so for now, it's being made slightly more expensive instead of being outright removed.
- Bionic PryArm - +6 - It was unnoticed for a while but, it could pry powered doors, so it's cost has been adjusted accordingly.

**Changelog**
:cl:
- tweak: Adjusted various trait point cost.
- fix: Sechud and Omnihuds can now be taken, one of the conditionals (department) wasn't working.
- fix: IPCs can no longer take Psionic Regeneration, not for the lack of trying but the lack of ability to process chemicals.

### DIFF
--- a/Resources/Prototypes/_DV/Traits/disabilities.yml
+++ b/Resources/Prototypes/_DV/Traits/disabilities.yml
@@ -21,7 +21,7 @@
   name: trait-poor-vision-name
   description: trait-poor-vision-desc
   category: DisabilitiesPhysical # Euphoria
-  cost: -1 #floof
+  cost: -3 #floof
   conditions:
   - !type:HasCompCondition
     component: Blindable
@@ -71,7 +71,7 @@
   name: trait-impaired-mobility-name
   description: trait-impaired-mobility-desc
   category: DisabilitiesPhysical # Euphoria
-  cost: -3 #floof
+  cost: -4 #floof
   effects:
   - !type:AddCompsEffect
     components:
@@ -84,7 +84,7 @@
   name: trait-amputee-left-arm-name
   description: trait-amputee-left-arm-desc
   category: DisabilitiesPhysical # Euphoria
-  cost: -3 #floof
+  cost: -5 #floof
   conflicts:
   - ArmAmputeeRight # TODO: AmputeeComponent doesn't stack, should be an effect instead
   effects:
@@ -99,7 +99,7 @@
   name: trait-amputee-right-arm-name
   description: trait-amputee-right-arm-desc
   category: DisabilitiesPhysical # Euphoria
-  cost: -3 #floof
+  cost: -5 #floof
   conflicts:
   - ArmAmputeeLeft
   effects:
@@ -125,7 +125,7 @@
   name: trait-painnumbness-name
   description: trait-painnumbness-desc
   category: DisabilitiesPhysical # Euphoria
-  cost: -2
+  cost: -3
   effects:
   - !type:AddCompsEffect
     components:
@@ -136,7 +136,7 @@
   name: trait-hemophilia-name
   description: trait-hemophilia-desc
   category: DisabilitiesPhysical # Euphoria
-  cost: -2 #floof
+  cost: -3 #floof
   effects:
   - !type:AddCompsEffect
     components:
--- a/Resources/Prototypes/_DV/Traits/disabilities.yml
+++ b/Resources/Prototypes/_DV/Traits/disabilities.yml
@@ -21,7 +21,7 @@
   name: trait-poor-vision-name
   description: trait-poor-vision-desc
   category: DisabilitiesPhysical # Euphoria
-  cost: -3 #floof
+  cost: -1 #floof
   conditions:
   - !type:HasCompCondition
     component: Blindable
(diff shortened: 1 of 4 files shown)


Cprn-Sprkl:
Reducing the cost of something because a species starts with it isn't great. The cost of species effects is playing those species.

Night vision is highly desirable and is one of the few techs I've been specifically asked to research by members of security. This skips the need for research, surgery, and materials for a highly desired effect. The cost should remain high for this one in particular. Not only to reflect the benefits it gives immediately at round start, but also to tax the loss in interactions for multiple departments.

Raikyr0:
> Reducing the cost of something because a species starts with it isn't great. The cost of species effects is playing those species.

The "Cost of Species" isn't balanced by default, all species have more positives then negatives. If we were to turn all the species quirks into traits, most will probably be in point debt by default (assuming there's 0 starting points).

ScarlettJFG:
I wouldn't say "All species", Humans don't really get anything extra. Slimes...well I did do my best to make sure they were Balanced, even with Geras (Looks at the three-four shots it takes from a squirt gun to knock a slime to crit). I can't speak for many species though I do know that Shadekin are unbalanced (More positives than negatives), as are Diona (I can think of only one positive frankly, the rest are neutral at best).


That all being said...I disagree with some of these, agree with most of these. I won't even bother arguing the ones I disagree with, you've already addressed them (Looks at psionic regen) and beating a dead horse is a waste of time, so I do think this is a net positive change, rather than a net negative.

Cprn-Sprkl:
> > Reducing the cost of something because a species starts with it isn't great. The cost of species effects is playing those species.
> 
> The "Cost of Species" isn't balanced by default, all species have more positives then negatives. If we were to turn all the species quirks into traits, most will probably be in point debt by default (assuming there's 0 starting points).

I'm not sure how this applies to what I said. If you want the benefits of a species, play that species. Species traits should have no bearing on how costly a trait is to take.

There is some argument to be had about re-visiting species due to the clear favouritism in the unique systems and amount of things certain species get while others are left with one impactful trait that is shared by three other species. That is separate from the discussion here though.

---------------------------

Round start cybernetics are already iffy due to skipping gameplay/interactions on a server that needs to stretch a one hour gameloop to three or four hours.  Every point cheaper means more people taking that route instead of engaging with the in game systems intended for those perks. They should cost quite a lot to encourage only those that consider them vital and defining will take them rather than leaning towards swiss army knife characters that start with everything just because they can.

I would not currently reduce a single point on any of the cybernetics, but as the absolutely most impactful of the eye variants I am actively arguing against any reduction in cost for the night vision eyes.

---------------------------

Outside of cybernetics the numbness trait doesn't have enough downside to support the very recent change to giving 2 points much less the additional point here. It has effects on surgery and is considered a small beneficial effect in devil contracts.

Honestly, overall this re-balancing seems aimed at increasing the overall pool of points available for people to min max with. The only change I outright agree with is the increase in psionic regen and prybar. Both of which I would prefer to see removed pending a rework, but more cost is better than nothing.

Raikyr0:
For Numbness, I was told the Health display doesn't change off good when hurt and has no bearing on surgery.

As for Nightvision, it's only labeled under cybernetics, it's intended to be more general, so people can have custom races that can see in the dark without it being cybernetic for their roleplay. 

In a perfect world, all the traits would be free, and people simply pick traits based on character roleplay, rather then gameplay being the determining factor. But it's not a perfect world where we can trust everyone to not make their character Adam Smasher.

kivdon:
I definitely agree with the changes to the disabilities, and mental. Ever since the rebase, combat has seemed so much deadlier, so it makes sense to give more point benefits to the things that decrease your combat ability or make it easier to be killed, like amputee and hemophilia and redshirt. 
For example, hemophilia feels a lot more dangerous on Euph than it did on Floof, so it makes sense to have it give an extra point.

Mnemotechnician:
Most of the cheapening changes are just straight up a "no" from me. We already have basically no way for a player to run out of points unless they are trying to get every upside available (which some players do powergame the shit out of, but that's not normal and shouldn't be endorsed), we don't need to make things cheaper.

SyaoranFox:
I am also saying no to this. This isn't fixing a balance issue. I would say that currently this PR creates imbalance. Feel free to do this again once we have more trait available for people to choose so that there actually is enough to make small deterministic changes.

Raikyr0:
I've readjusted it.

SyaoranFox:
Thank you for making these adjustments, but currently I don't think its worth anyone's time rebalancing the costs of things when there are so few choices to begin with. If you want to redo this once we have more trait options available, then we can talk about the merits and costs of each trait.

PR #493, opened 2026-04-29
Adjusted Meteors

Raikyr0:
## About the PR
Halves the explosive strength of meteors

ALRIGHT I'VE DONE IT.
I've nerfed space dust to a point where it won't break walls but will still blow reinforced windows.
To compensate I've increased the amount of dust per storm.

## Why / Balance
Meteors have been causing way too much damage too often, and Reinforced Walls still crumple to space dust.

**Changelog**
:cl:
- tweak: Changed Meteors less frequent on greenshifts.
- tweak: Changed Greenshift meteors to only be dust storms.
- tweak: Changed Meteors to be less destructive.

[Classic Video too big to attach](https://cdn.discordapp.com/attachments/1255992182880473269/1499938973928390787/2026-05-02_10-41-01.mp4?ex=69f69e86&is=69f54d06&hm=e1453175adb892c7575b374dd85428b23acbd85539fde89843ffca0e9e88582a&)

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/meteors.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/meteors.yml
@@ -68,7 +68,7 @@
         - Impassable
         - BulletImpassable
   - type: Explosive
-    totalIntensity: 25
+    totalIntensity: 12.5 # Euphoria - Halved
   - type: Destructible
     thresholds:
     - trigger:
@@ -103,7 +103,7 @@
         - Impassable
         - BulletImpassable
   - type: Explosive
-    totalIntensity: 100
+    totalIntensity: 50 # Euphoria - Halved
   - type: Destructible
     thresholds:
     - trigger:
@@ -143,7 +143,7 @@
         - Impassable
         - BulletImpassable
   - type: Explosive
-    totalIntensity: 200
+    totalIntensity: 100 # Euphoria - Halved
   - type: Destructible
     thresholds:
     - trigger:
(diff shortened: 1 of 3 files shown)


Mnemotechnician:
I'm not sure about this, can you show an example of these changes in-game?

TheRedForestButcher:
I believe a better change would be to lower the intervals of meteors. My experience in engineering for over 1K hrs alone on this server I feel as if meteors happen so often and it feels like the only event that gets triggered. 

Meteor damages are fine and pretty easy to repair, I feel like meteor events are just too common.

Raikyr0:
> I'm not sure about this, can you show an example of these changes in-game?

I'll record a video of it, assuming I can get it to actually hit the station in dev

Mnemotechnician:
> > I'm not sure about this, can you show an example of these changes in-game?
> 
> I'll record a video of it, assuming I can get it to actually hit the station in dev

meteor swarm rule targets random points within the station while trying to avoid protected grids. The targetting is centered around the observer spawnpoint

TheLunaticOfTheDrunkenMoon:
> I believe a better change would be to lower the intervals of meteors. My experience in engineering for over 1K hrs alone on this server I feel as if meteors happen so often and it feels like the only event that gets triggered.
> 
> Meteor damages are fine and pretty easy to repair, I feel like meteor events are just too common.

I think if the interval is changed all this will do is we will be barraged by a bunch of other events instead due to how the scheduler currently works? I'm not fully certain on this, but it feels.... weird .... especially when you are at evac and you get 5 events back to back and causes all the alerts to combine into one

SyaoranFox:
Due to having no approvals, I am closing this PR. I recommend that you provide as much evidence as you can as part of your PRs. It can be difficult to accept changes when they are so inherently random.

PR #495, opened 2026-04-29
Adds Rat Servants to Trash Bag Whitelist

Raikyr0:
## About the PR
Adds Rat Servants to Trash Bag Whitelist so you can suck em all up after a rat massacre.

**Changelog**
:cl:
- add: Added Rat Servants to Trash Bag whitelist

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/NPCs/regalrat.yml
+++ b/Resources/Prototypes/Entities/Mobs/NPCs/regalrat.yml
@@ -314,6 +314,7 @@
     tags:
       - CannotSuicide
       - FootstepSound
+      - Trash # Euphoria - QoL
   - type: NoSlip
   - type: MobPrice
     price: 500 # rat wealth


Mnemotechnician (approved):

Mnemotechnician:
Although this has the unintended consequence of letting you trashbag live rat, which is not really desirable.

SyaoranFox:
I can't approve of this. Trashbags would become a meta choice against a rat king with lots of kids. Literally you could use the mass area collect and force the mob into the bag who won't try to escape.

Mnemotechnician (changes requested):
Yeah, nevermind, this is a bad solution. You could code a system that tags mobs as trash when they die (there's already such a system but for drinks and the like). Maybe there's already one, idk.

Raikyr0:
So, this change isn't a new change, it's one that we used to have on the old codebase. What used to happen was that the rat servants when set to loose would instantly kill the person holding the bag, this also includes rat servants in bags and pockets.

SyaoranFox:
> So, this change isn't a new change, it's one that we used to have on the old codebase. What used to happen was that the rat servants when set to loose would instantly kill the person holding the bag, this also includes rat servants in bags and pockets.

Yes. I have seen people use trashbags as a solution to a rat king. Filling the bag then throwing it away from them. I stand by my thoughts.

PR #498, opened 2026-04-30
Content Warning: Vore Patch 2

JaDiDo:
## About the PR

Reworked the voresystem and component plus new feature of handling verbs inside containers

## Why / Balance

Massive Reduction of who gets the components, increasing code longevity, better handling of storage

## Technical details

1. VoreComponent will now be assigned during both consent changes and startup
-> meaning no more getting that component if you have neither prey or pred on!
-> small timer function needed to get the most recent values
2. Renamed a couple of methods based on if they are subscription (On[...]), or not
-> small change but should now make code slightly easier to read in my opinion
3. Included radiation immunity
-> based on the component function its potassium iodine
-> sadly only gives 90 percent rad immunity cause of dv but should give prey time to react and act
4. RemoveStomach method now checks if someone is still in a vore container
-> in case a prey gets released but still ends up inside another pred such as during multi vore
-> small timer function needed to get the most recent values
5. Vore inside a container is now possible (verbs show up) 
-> only works however if both entities involved are inside the same container
-> else method will just return
-> ensureentitymanager should prevent issues by ejecting them from the initial container
6. Reworked vorecomponents
-> IntentionalRelease doesn't really need a Datafield since its shortlived
-> vore container is now a var inside with all the references patched which will be useful later for multi containers
7. Helper methods added
-> Waitforupdate to create recent values (else for example consent will use old values)
-> IsInVoreContainer (line has been used a lot so now its put there for code readability)
-> IsValidVoreInteraction to check if both preys are in the same vore container
8. During consent check the prey can be ejected of a container if they turned off prey toggle
-> acts as a emergency leave (is inside applyvoreconsent as it handles the change from prey to notprey)

## Media
Screenshot 2026-04-29 190728

</p></details>

## Requirements

## Licensing:

**Changelog**

:cl:
- add: Patched handling components with voresystem

### DIFF
--- a/Content.Server/_Floof/Vore/VoreSystem.cs
+++ b/Content.Server/_Floof/Vore/VoreSystem.cs
@@ -30,7 +30,6 @@ public sealed class VoreSystem : EntitySystem
     [Dependency] private readonly SharedPopupSystem _popupSystem = default!;
     [Dependency] private readonly IConfigurationManager _cfg = default!;
     [Dependency] private readonly CarryingSystem _carryingSystem = default!;
-    [Dependency] private readonly ITimerManager _timer = default!;
 
     public static readonly ProtoId<ConsentTogglePrototype> isPred = "PredVore";
     public static readonly ProtoId<ConsentTogglePrototype> isPrey = "PreyVore";
@@ -186,7 +185,7 @@ public sealed class VoreSystem : EntitySystem
             return;
 
         var pred = uid;
-        var container = _containerSystem.EnsureContainer<Container>(args.User, comp.ContainerId);
+        var container = _containerSystem.EnsureContainer<Container>(pred, comp.ContainerId);
         
         var count = 0;
         //only counts entities with bodies meaning no items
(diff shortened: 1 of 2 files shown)


Content.Server/_Floof/Vore/VoreSystem.cs
```diff
@@ -288,21 +325,83 @@ should prevent possible exploitation of the system*/
...
+        });
+    }
+
+    /// <summary>
+    /// used to delay certain actions by a tick to make sure all the recent changes have been applied 
+    /// for example such as consent changes or container changes, else will often use old values
+    /// </summary>
+    private void WaitForUpdates(Action action){
+        Timer.Spawn(0, action);
+    }
```
Mnemotechnician: This is awful, don't use timers ever. You're asking RobustToolbox to give you a race condition. Instead, store a list of things that need updating and handle them in EntitySystem.Update

Mnemotechnician (changes requested):

Content.Server/_Floof/Vore/VoreSystem.cs
JaDiDo: Ah sorry about that! Have reworked the method alltogether into update

Mnemotechnician (approved):
looks fine to me code-wise

M3739 (approved):

SyaoranFox (approved):

PR #509, opened 2026-05-02
Make "spawn here" verb load traits and loadouts

Mnemotechnician:
## About the PR
This sucks. Only loads passenger gear and traits cus there's no reasonable way to prompt the admin for a job when using verbs.

## Media

https://github.com/user-attachments/assets/5ebc24a0-39d9-41c6-a477-ddd0784ad044

</p></details>

**Changelog**
:cl:
- fix: The "spawn here" admin verb now loads the loadouts & traits from the player's profile (while assuming Passenger as the job).

### DIFF
--- a/Content.Server/Administration/Systems/AdminVerbSystem.cs
+++ b/Content.Server/Administration/Systems/AdminVerbSystem.cs
@@ -157,8 +157,15 @@ namespace Content.Server.Administration.Systems
 
                             var stationUid = _stations.GetOwningStation(args.Target);
 
+                            var job = "Passenger"; // Floofstation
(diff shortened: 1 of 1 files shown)


Cprn-Sprkl:
Would it be possible to have a pop up once the verb is executed that is populated with a list of job prototypes?

Something like this one:
image

Otherwise would these changes be something we could mirror on to the "spawncharacter" command? That one take a job as an argument.

Mnemotechnician:
Unfortunately that's outside the scope of this command, the primary scope is making it load traits. Jobs dont really affect anything other than starting clothing when using spawncharacter/the verb.

PR #510, opened 2026-05-03
Increase alcohol vomiting threshold

Raikyr0:
## About the PR
~~Reduces the "BoozePower" of Ethanol from 2 to 1, allowing people to drink more before their vision is blurred.
This does not change the damage threshold.~~

Edit 2026-08-13: now this PR only doubles the vomiting threshold.

## Why / Balance
~~So, I heard some complains about slimes being able to drink an Op's Shine before the rebase without getting drunk, but it makes em drunk in 1 glass now. But nothing's really changed other then some sort of DV calculation which is impossible to find in the C# code. So, instead of fucking with C#, I've halved the "Drunkness" buildup so people are less lightweights, foreseen consequences will probably be more alcohol poisoning tho. ~~

**Changelog**
:cl:
- tweak: Doubled the vomiting threshold from alcohol, so no more getting sloshed in 1 glass, now it's 2 glasses.

### DIFF
--- a/Resources/Prototypes/Reagents/Consumable/Drink/alcohol.yml
+++ b/Resources/Prototypes/Reagents/Consumable/Drink/alcohol.yml
@@ -207,7 +207,7 @@
           type: [ Dwarf ]
           inverted: true
       - !type:Drunk
-        boozePower: 2
+        boozePower: 1 # Euphoria - People getting drunk too fast
 
 - type: reagent
   id: Gin
--- a/Resources/Prototypes/Reagents/Consumable/Drink/alcohol.yml
+++ b/Resources/Prototypes/Reagents/Consumable/Drink/alcohol.yml
@@ -201,13 +201,13 @@
         conditions:
         - !type:ReagentCondition
           reagent: Ethanol
-          min: 12
+          min: 25
         # dwarves immune to vomiting from alcohol
         - !type:MetabolizerTypeCondition
           type: [ Dwarf ]
           inverted: true
       - !type:Drunk
-        boozePower: 1 # Euphoria - People getting drunk too fast
+        boozePower: 2
 
 - type: reagent
   id: Gin


ScarlettJFG:
Rephrasing what I said when I mentioned it: It got them drunk, but it didn't damage them. It did also seem to get them less drunk but not 2 glasses drunk.

Basically I would have about half the drunkeness I get off of one now, and no damage as a slime.

(really should have worded shit better)

ScarlettJFG:
Was asked about this tonight. Took a look in the floofbase, code is the same as DV so this would change everyone and cause problems outside of slimes/Ops. Digging a *bit* deeper into this.

Resources/Prototypes/Reagents/Consumable/Drink/alcohol.yml
ScarlettJFG: Change this to 25 to reflect Floofbase

Resources/Prototypes/Reagents/Consumable/Drink/alcohol.yml
```diff
@@ -207,7 +207,7 @@
           type: [ Dwarf ]
           inverted: true
       - !type:Drunk
-        boozePower: 2
```
ScarlettJFG, line 210: Revert this

ScarlettJFG (changes requested):
Ok after digging around a bit at request, these changes will reflect parity with floof base, and fix the issue at hand. Ops' Shine (the drink that brought this up) causes a maximum of 22.98 ethanol to be in the system from one singular glass, meaning you can even still take another few u of Ops immediately afterwards (around three by my calculations) and still not hit the 25u threshold for vomit...except it's 12u ethanol in our current code.

SyaoranFox (approved):

Mnemotechnician:
Big no. If you set the booze power of ethanol to 1, it will prevent the Drunk status effect from building up at all.

Mnemotechnician:
Okay, the PR contents no longer match the PR name or description. Let me edit them.

Mnemotechnician (approved):
Please make sure to update the description when changing course like that.

PR #524, opened 2026-05-05
Banana cream Pie

Raikyr0:
## About the PR
Adds Reagent Banana Cream
Replaces Banana Cream Pie's Reagents with Banana Cream

## Why / Balance
So, when someone gets hit by a banana cream pie currently.
They leave a pile of green nutriment, and that doesn't feel very pie like.
So I've created a new chemical that goes into the banana cream pie (the only pie that can be thrown) that is the color of banana cream.

No more green flood from pies, it will now be banana cream.

image

## Technical details
First I copy Nutriment, place it in our namespace so there will be no issues later.
Then I create a replacement to nutriment with a tiny bit of vitamins folded in.
Write flavor (banana) text for the pie.
Then I add SolutionContainerManager to the pie itself, to overwrite the original solution. As well as the pie slice.

**Changelog**
:cl:
- tweak: Changed the chemicals inside a Banana Cream Pie, it will now look tasty instead of green.


SyaoranFox:
Did you investigate what would happen if you parented nutriment and changed the colour? I don't mind the healing, but I am thinking we can cut down on the code duplication.

SyaoranFox (approved):

Mnemotechnician (approved):

PR #525, opened 2026-05-05
Ghost Role: Sample Collector

Cprn-Sprkl:
## About the PR

Adds the Sample Collector unknown shuttle to the unknown shuttle event pool, and the Sample Collector ghost role. This new ghost role spawns as the player's selected character.

Also includes a system so that ghost roles that spawn a player's character will also apply their traits.

## Why / Balance

This ghost role offers avenues for RP and income for station crew. Includes a new shuttle for a new location to RP.

## Technical details

The big change is an addition to TraitSystem that listens for ghost spawning and applies traits to ghost roles that spawn a player's character.

## Media

image
image
image

</p></details>

## Requirements

## Licensing:

## Breaking changes

N/A - Tested a few non-player character ghost roles and they seem to spawn fine.

**Changelog**

:cl:
- add: Sample Collectors have arrived to service station crew.

### DIFF
--- a/Content.Server/Administration/Systems/AdminVerbSystem.cs
+++ b/Content.Server/Administration/Systems/AdminVerbSystem.cs
@@ -157,8 +157,15 @@ namespace Content.Server.Administration.Systems
 
                             var stationUid = _stations.GetOwningStation(args.Target);
 
+                            var job = "Passenger"; // Floofstation
                             var profile = _gameTicker.GetPlayerProfile(targetActor.PlayerSession);
-                            var mobUid = _spawning.SpawnPlayerMob(coords.Value, null, profile, stationUid);
+                            var mobUid = _spawning.SpawnPlayerMob(coords.Value, job, profile, stationUid);
+
+                            // Floofstation - raise playerspawcomplete so traits and the like can load.
+                            var targetSession = targetActor.PlayerSession; // Note: PlayerSession gets set to null after TransferTo. Mind the side-effects.
+                            RaiseLocalEvent(mobUid,
+                                new PlayerSpawnCompleteEvent(mobUid, targetSession, job, true, true, 0, stationUid ?? EntityUid.Invalid, profile),
+                                true);
 
                             if (_mindSystem.TryGetMind(args.Target, out var mindId, out var mindComp))
                                 _mindSystem.TransferTo(mindId, mobUid, true, mind: mindComp);
--- a/Content.Server/Administration/Systems/AdminVerbSystem.cs
+++ b/Content.Server/Administration/Systems/AdminVerbSystem.cs
@@ -157,15 +157,8 @@ namespace Content.Server.Administration.Systems
 
                             var stationUid = _stations.GetOwningStation(args.Target);
 
-                            var job = "Passenger"; // Floofstation
                             var profile = _gameTicker.GetPlayerProfile(targetActor.PlayerSession);
-                            var mobUid = _spawning.SpawnPlayerMob(coords.Value, job, profile, stationUid);
-
-                            // Floofstation - raise playerspawcomplete so traits and the like can load.
-                            var targetSession = targetActor.PlayerSession; // Note: PlayerSession gets set to null after TransferTo. Mind the side-effects.
-                            RaiseLocalEvent(mobUid,
-                                new PlayerSpawnCompleteEvent(mobUid, targetSession, job, true, true, 0, stationUid ?? EntityUid.Invalid, profile),
-                                true);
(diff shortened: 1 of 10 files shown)


Dunrab (changes requested):
A few small things in regards to the shuttle:

- Did you purposely use a debug generator? If so I worry will players get wind of this and steal the DebugGenerator for other use such as infinite power on station
- On any lattice you have exposed such as the ones for thrusters you will want to place Atmos Fix Vacuum Marker (AtmosFixBlockerMarker)

Optional changes (not required, just some QOL for the ghostrole taker):

- The shuttles waste line uses a passive vent, which will work on all our maps but for our planet side maps, that means the passive vent will currently take in atmos so you may want to swap to a air injector
- You don't have to do this but you can if you would like you can use GasMixerEnabled to have the shuttle distro automatically come on when the shuttle is spawned in
- Under the docking airlock i see you placed a regular tinyfan which of course will prevent the shuttle from loosing air, but that will tile the tiny fan is on will loose pressure over time from players going in and out or just typing while they stand on the tile, i suggest changing it to AtmosDeviceFanDirectional so the tile doesn't have that issue
- Also not required, just more so if you wanted to, you can give the shuttle the IFFComponent if you wanted to change the color of shuttles display name on a mass scanner or shuttle console.
- Perhaps a medical record PC? I am not sure if its really needed per say but it may lead to some folks RPing out that they are updating players medical records after a donation

Other than that, everything else looks good to me.

Cprn-Sprkl:
I'll start working on some of these. The debug generator was used to address a complaint I kept seeing for all ship based ghost roles. The material fed generators all burn through fuel incredibly fast and leave them without any power. That's really bad for roles that take a lot of set up, like the syndicate chemist, or roles that might be occupied for extended periods of time, like the sample collectors. I'll see about making it not able to be unanchored and anyone circumventing that to steal it for other use would be smacked for power gaming.

I am open to other suggestions if there is a compact source of power that is more efficient than the pacman generators or even less tempting for people to try and take.

Dunrab:
> I'll start working on some of these. The debug generator was used to address a complaint I kept seeing for all ship based ghost roles. The material fed generators all burn through fuel incredibly fast and leave them without any power. That's really bad for roles that take a lot of set up, like the syndicate chemist, or roles that might be occupied for extended periods of time, like the sample collectors. I'll see about making it not able to be unanchored and anyone circumventing that to steal it for other use would be smacked for power gaming.
> 
> I am open to other suggestions if there is a compact source of power that is more efficient than the pacman generators or even less tempting for people to try and take.

Yeah no I totally agree in that regard, we need to go through all the event shuttles at some point and give proper power sources that do not rely on being fed plasma or uranium. 

You could use a mix of regular 15kw generators with a shuttle APU or two and then make sure it has enough power that way, it shouldn't need more than 20Kw hopefully, if that isn't enough you could make the debug one unanchorable by changing the comp on it but then there is also an admin burden of checking to see if someone stole it using an RCD to destroy the floor each round it pops up in

Cprn-Sprkl:
image
Engineering is a bit tight, but I'm thinking of moving the APC and air alarm to either side of the door and place a second 15kw gen where the toolbox is now. Basically three RTGs in the space of two.

Cprn-Sprkl:
image
I've hit everything on the list, but I noticed something. The doors changed their paint. Is there a specific method or tool other than the spray painter to get the doors set up with the right access AND the look I want?

I had already repainted them in this picture.

Dunrab:
> image I've hit everything on the list, but I noticed something. The doors changed their paint. Is there a specific method or tool other than the spray painter to get the doors set up with the right access AND the look I want?
> I had already repainted them in this picture.

Awesome. As far as door access goes, you could try to use an access config on it, save the grid then load it onto an inited map to see if it properly saves since I am unsure, and I am also unsure how to paint a door, you may be able to add the components and set them but that is something I would have to ask @Mnemotechnician because I don't know if that would even work

Cprn-Sprkl:
Unfortunately from the first iterations I know the access config stuff does not save. For now I might just slip a spray painter and a "professionalism preparation" sheet of instructions into the cockpit so the crew can at least get rid of the ugly public access door paint before they ftl onto the station map.

Cprn-Sprkl:
image
Changes all set, including a requested cigarette vendor.

Dunrab:
> Unfortunately from the first iterations I know the access config stuff does not save. For now I might just slip a spray painter and a "professionalism preparation" sheet of instructions into the cockpit so the crew can at least get rid of the ugly public access door paint before they ftl onto the station map.

You could always make a shuttle docking airlock with the access you want it to have, but ultimately not required

Dunrab (approved):
Shuttle looks good to go, will be nice to have more ghost roles like this

PR #529, opened 2026-05-05
Scarf

Raikyr0:
## About the PR
Delta V added a scarf category and a bunch of other stuff with only scarfs.
Unfortunately for us, we already had scarfs.

This PR
Removes the scarf category and several redundant ones.
Replaced redundant job ones with departmental categories.
Adds said categories to jobs. 
Sorts the original job loadout (ie moved librarian and chaplain to science etc)
Adds missing scarf colors.

## Why / Balance
Bothered me.

## Breaking changes
Upstream PRs that add loadout stuff may break in the re-sorted base job loadout.

**Changelog**
:cl:
- add: Added missing scarf colors from loadout.
- remove: Removed DV's scarf tab.
- add: Added departmental neck tabs to replace several tabs that was only scarfs.


Raikyr0:
the last fucking test fail was a fluke

PR #532, opened 2026-05-06
Bible Delay Fix

Cprn-Sprkl:
## About the PR

Fixes the use delay on the bible.

## Why / Balance

Prevents the unintended spamming of bible healing.

## Technical details

A single line.

## Media
N/A

## Requirements

## Licensing:

## Breaking changes

N/A

**Changelog**

:cl: sprkl
- fix: Fixed an issue with bibles.


Mnemotechnician (approved):
Both branches of this if-statement do the exact same thing, weird. A better solution would be to call TryResetDelay in the beginning (instead of IsDelayed), but this is fine.

PR #535, opened 2026-05-07
Mail Tracker

Cprn-Sprkl:
## About the PR

Adds the mail tracker, a special pinpointer that can scan mail and point to the recipient.

## Why / Balance

A fun gadget for all mail carriers, and a confidence booster for those that want to deliver mail, but don't know all the map layouts very well.

## Technical details

Adds functionality to the mail system to detect when mail is tapped on by a pinpointer and set the target if that pinpointer is allowed to scan mail.
Adds functionality to the pinpointer system and component to identify if a pinpointer should be able to scan mail.

## Media

image
image
image
image
image
image
image
image

</p></details>

## Requirements

## Licensing:

## Breaking changes

N/A

**Changelog**

:cl:
- add: Added a Mail Tracker to the Courier for finding mail recipients.


kivdon:
Is this something that couriers will be able to spawn with in their loadout? 

And are there any limitations to its usage?

ScarlettJFG:
Blacklist the Captain! We do not need a poor man's Disk pinpointer.

TheRedForestButcher:
I feel like could be more “balanced” if it only worked if the person had suit sensors on and on a valid grid. 

This could be cheesed to find people deep space or trying to hide otherwise would be impossible to find. 

Also people could just save their old mail so it could be like a meta pinpointer if they get lost etc…

Good idea I just believe it *may* cause issues.

kivdon:
> I feel like could be more “balanced” if it only worked if the person had suit sensors on and on a valid grid.
> 
> This could be cheesed to find people deep space or trying to hide otherwise would be impossible to find.
> 
> Also people could just save their old mail so it could be like a meta pinpointer if they get lost etc…
> 
> Good idea I just believe it _may_ cause issues.

There is potential for it to be misused, but at such a low possibility...
It's based on having received mail. So the chance of mail coming in for someone you want to find is already very low. Now let's say that mail did come in: What are the chances that you intercepted that mail? If you're not a courier, very low. And what are the chances that the person is hiding? Also very low. Multiplying three low chances together makes an exponentially low chance.

Basically, I'm saying that while it's absolutely valid to point out that this could be used for finding people in ways outside of the intended purpose, the actual chance of such an opportunity occurring is absolutely miniscule, especially when compared to how useful it is for its intended use. 

I definitely **would not** have it be only valid for those with suit sensors on, because then you might as well not have it at all. Just use a crew monitor at that point, and then you get a whole map and everything. I believe one of the stations even has a crew monitor computer in the mail room for that reason. But as a courier main, I can tell you this pinpointer would be a lot more fun (and useful) than a crew monitor. 

I'll also say that Afterlight recently implemented something similar, except it's even more powerful since it shows you directly on the map where the mail target is. As far as I've seen, this has not been an issue there.

Mnemotechnician (changes requested):
This allows indefinite and non-counterable cross-grid tracking of any person who's ever happened to receive mail. This could be used by antags and security to track down people with a level of precision not ever seen in ss14. I'm against this without some significant changes (at the very least requiring opt-in and linking to an ID card and not to the recipient)

Mnemotechnician:
Or better yet, this should rely on the suit sensor system and not on the pinpointer system (because, once again, pinpointers are free cross-grid tracking)

Cprn-Sprkl:
There. It can no longer be researched or built in any way.

The only way one shows up on shift is in the pocket of an arriving courier. If an antag is worried about it being used against them they can opt to steal it first. The stars would have to align that the courier is willing to help track the antag and that they have an unopened piece of mail for the specific antag ready to go.

It can only start tracking on locked mail. Soon as you tap your ID that piece of mail can't start a new tracker so no digging through the trash for open mail to track. A courier would have to horde mail for everyone in hopes of having a chance to track someone and at that point they should probably be fired for costing cargo money and not doing their job.

As for antag using this. Is it that big a deal? This just gives antag couriers a toy to play with and won't make a significant outcome in whatever they were going to do. There aren't really many scenarios where crew are in active hiding from traitors.

TheLunaticOfTheDrunkenMoon:
I'd like to point out, it only costs logi money if you don't do the urgent ones, and unless you don't do *any* of them, there's a good chance your LO or anyone monitoring doesn't take it into mind or consideration unless you are leaving all the mail laying around in the mail room.... with it not being passengers that gets a pass alot of the time or if you say 'unless it's urgent, they have to come to me'.... but if you put it in a bag and suddenly they don't complain anymore... or you turn of the mail and suddenly it seems like you are on break instead.... You can still very much hoard the mail you want, like the various heads so you can sell your services to do that but.... it's also just a *free* crew monitor that targets one person with accuracy


The things I want to ask, what happens if you put a knife to the mail? Would that remove the current tracker? What if the person you are tracking goes to cryo before you can reach them? Do you then have to bother a head to get their id to commit mail fraud so you can still use your tracker? What happens if you don't have a head and you aren't an antagonist and this happens???

Cprn-Sprkl:
> What if the person you are tracking goes to cryo before you can reach them? Do you then have to bother a head to get their id to commit mail fraud so you can still use your tracker? What happens if you don't have a head and you aren't an antagonist and this happens???

You just tap a different piece of mail.

TheLunaticOfTheDrunkenMoon:
> > What if the person you are tracking goes to cryo before you can reach them? Do you then have to bother a head to get their id to commit mail fraud so you can still use your tracker? What happens if you don't have a head and you aren't an antagonist and this happens???
> 
> You just tap a different piece of mail.

"Soon as you tap your ID that piece of mail can't start a new tracker" is why I was confused, and misinterpreted trash and thought you meant a hoard of undelivered mail. So I then mistook everything as you could only have one permanently active mail tracker that you are adamant on delivering to one specific person that requires it to potentially be opened or damaged before you can try someone else. My mistake lmao

kivdon:
> Or better yet, this should rely on the suit sensor system and not on the pinpointer system (because, once again, pinpointers are free cross-grid tracking)

As I said in my comment, this absolutely **should not** rely on the suit sensors being on, because then you might as well just be using the crew monitor, and we lose the benefits of what this is intended for.

And as has been said, the stars would have to align with so many low chances coming together for this to actually be of any use for purposes outside of the intended purpose. It's valid and good to point out the ways things could _potentially_ go wrong, but the likelihood of those things should be compared against the benefit of the system. This thing is waaaaaaaay more useful to an actual good-intentioned courier than to an antag. 

For this to be useful outside of the intended purpose, the following things must **ALL** be true:
The target is able to receive mail, and in hiding. (low chance)
and
Mail was generated for the target. (low chance)
and
You have the mail. (low chance if you're not the courier)
and
You have the pinpointer. (low chance if you're not the courier)

For this to be useful for the intended purpose, _any_ of the following things can be true:
The courier does not have access to a crew monitor (high chance).
or
The courier has access to a crew monitor, but the target doesn't have coords on (medium chance).
or
The target is not responsive on the radio for trying to get their location for mail delivery (high chance). 

This thing is useful and fun for its intended purpose, and exceedingly unlikely to be useful outside of its intended purpose.

TheLunaticOfTheDrunkenMoon:
Chances are only minuscule if you do take in *all* the odds and run them as "everyone in the lobby has antag enabled or otherwise". I think it should be based on suit sensors *because* someone's going to roll antag, and get found out by the mailman cause they in the *unlikely* scenario got found whilst taking every precaution to not get caught by someone that's not part of command, security, or the AI... and they immediately go blab or report it in..... Or worse, gives the pinpointer to security to catch on-the-run criminals. And that bit is what makes it worse cause all you have to do is stalk or ask the mailman and suddenly the potential of having your antag roll ruined exists without any counterplay

Cprn-Sprkl:
If security already knows the exact name of the person they're looking for that antag has already messed up. If the stars align to make a situation where an antag has messed up, is on the run, has unopened mail in their name, a courier is present, and the courier is willing to help then yes, this will have an impact on an antag's round.

Some things I'd be willing to do:

- Change it so it points to the ID card of the person instead of the person.
- See about adding an access reader so only people with mail access can use it.

If making it rely on the sensor system is a hard line for the maintainers then hopefully two straight up say that here so I can stop wasting time.

SyaoranFox:
> If security already knows the exact name of the person they're looking for that antag has already messed up. If the stars align to make a situation where an antag has messed up, is on the run, has unopened mail in their name, a courier is present, and the courier is willing to help then yes, this will have an impact on an antag's round.
> 
> Some things I'd be willing to do:
> 
> * Change it so it points to the ID card of the person instead of the person.
> * See about adding an access reader so only people with mail access can use it.
> 
> If making it rely on the sensor system is a hard line for the maintainers then hopefully two straight up say that here so I can stop wasting time.

I think with both of those changes, the tracker would be acceptable, and functional towards the problem being solved. All jobs should have some kind of a perk, and courier didn't really have one. I believe this fits that.

I also want to remind people that they shouldn't argue here. You can give an opinion on a PR, but debate, especially with people who are not working on the PR, will be discouraged. Debate shouldn't be a barrier for a PR to get through, just the merits of its balance, and effect on the game itself.

Cprn-Sprkl:
Mail tracker now requires mail access to use. It also now tracks the ID instead of the person. Side bonus, it will fail to start tracking if the target is not wearing their ID.

SyaoranFox (approved):
I am interested to see how this is used.

Cprn-Sprkl:
I added polish to the tracker. It now displays pop up messages and plays sound effects to notify the user of outcomes. Different things based on finding the target, being unable to find the target, and not having the access to use it. A use delay has been set up so the noises can't be spammed.

philippaz272006-commits (approved):

PR #560, opened 2026-05-13
Lewd access toggle and lewd system improvements

Mnemotechnician:
## About the PR
God damn it, i pushed to the wrong repo.

This adds a new interaction verb - clothing bypass - that lets you toggle whether your lewd organs can be used and interacted with while bypassing (ignoring) clothing.

Also improves the slot blocker system to allow using "virtual" (not really virtual, but spawned in nullspace) clothing to check slot obstruction. This means that if you wear e.g. a non-blocking clothing piece such as a winter coat and nothing underneath, it won't stop you and others from interacting with undies and such. Currently this is only used in the undies system and in the lewd interaction verbs.

Also I realized that the LewdInteractions consent wasn't used at all. I removed the toggle proto for now. It doesn't add much and only leads to confusion.

## Media

Showcase of the new slotblocker behavior

https://github.com/user-attachments/assets/a2d31130-687a-4740-bfda-e3c0a05482c9

Showcase of the bypass toggle and lewd interaction verbs working if the target has just an outer clothing with nothing underneath

https://github.com/user-attachments/assets/59c21a06-2c43-4e5f-89f9-bca39726c073

God fucking damn it, and Spectacle encoded them in vp8 which github doesn't support. Pain.

</p></details>

**Changelog**
:cl:
- add: There is now a "clothing bypass" toggle that you can access by right-clicking yourself. Toggling it on allows others and you to interact with your lewd organs regardless of clothing.
- tweak: Some verbs like "toggle undies" can now be used even if the target has outer clothing, so long as they don't have anything underneath and the outer clothing is not a hardsuit or similar.
- fix: Removed the redundant "lewd interactions" consent toggle as it wasn't used anywhere.
- fix: The ai eye should no longer be able to see interaction verb popups from anywhere near a camera.

### DIFF
--- a/Content.Server/_Floof/InteractionVerbs/Actions/Lewd/ToggleLewdAccessibleAction.cs
+++ b/Content.Server/_Floof/InteractionVerbs/Actions/Lewd/ToggleLewdAccessibleAction.cs
@@ -1,0 +1,25 @@
+using Content.Shared._Floof.InteractionVerbs;
+using Content.Shared._Floof.Lewd.Components;
+
+namespace Content.Server._Floof.InteractionVerbs.Actions.Lewd;
+
+/// <summary>
+///     Toggles whether others' and the user can interact with their lewd organs through clothing.
+/// </summary>
+public sealed partial class ToggleLewdAccessBypassAction : InteractionAction
+{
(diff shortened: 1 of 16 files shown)


Mnemotechnician:
Gonna merge this now, it *should be fine*™

PR #569, opened 2026-05-15
Make ModifyLewdOrgansEffect increase solution volume to accomodate for production

Mnemotechnician:
## About the PR
It used not to do that, which caused "quadruple production" and similar traits to have no effect on organs that had no extra volume on top of what's needed for production.

## Technical details
Web-edited, but it should be fine.

## Media
<details>
Before:

image
image

After:
image
image

</details>

**Changelog**
:cl:
- fix: Lewd traits that increase the maximum amount of fluid production should now correctly increase the organ fluid capacity to accommodate for additional fluids.


Mnemotechnician:
anyway, it works (see media)

PR #571, opened 2026-05-15
Double Barrel Shotgun - Both Barrels

Raikyr0:
## About the PR
Adds a burst fire option to fire both barrels of shotguns with 2 barrels.
Ported from Foxtrot where it was tested, some people also asked for it.

## Why / Balance
Because shotguns with 2 barrels are designed to be able to fire both at once.

## Requirements

**Changelog**
:cl:
- tweak: Allows Shotguns with 2 Barrels to unload both in a single click.


SyaoranFox (approved):
I am cautious about these changes because it does x2 the damage with this mode. We will need to watch to see if this becomes a problem.

PR #577, opened 2026-05-15
Trait System: Character Swap Bug Fix

Cprn-Sprkl:
## About the PR

Fixes the issue where failed trait conditions on one character blocked the trait from selecting when swapping to another character. An example of the bug: having night vision eyes on a diona then swapping to a tajaran, then swapping back would not re-select the night vision eyes trait.

## Why / Balance

General bug fix

## Technical details

Created a set of commands to clear all the conditions from every trait. This is run before applying the new traits and then validating them with the new set of conditions.

## Requirements

## Licensing:

## Breaking changes

N/A

**Changelog**

:cl: sprkl
- fix: Fixed traits not applying correctly when swapping characters.

### DIFF
--- a/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
+++ b/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
@@ -588,8 +588,8 @@ namespace Content.Client.Lobby.UI
                 }
             }
 
-            Traits.SetSelectedTraits(selectedTraits);
             Traits.UpdateConditions(Profile);
+            Traits.SetSelectedTraits(selectedTraits);
         }
         // End DeltaV - Traits Integration
 
--- a/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
+++ b/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
@@ -588,8 +588,9 @@ namespace Content.Client.Lobby.UI
                 }
             }
 
-            Traits.UpdateConditions(Profile);
+            Traits.ResetConditions(); //Euphoria | Clears previous profile's conditions.
             Traits.SetSelectedTraits(selectedTraits);
+            Traits.UpdateConditions(Profile);
         }
         // End DeltaV - Traits Integration
 
--- a/Content.Client/_DV/Traits/UI/TraitCategory.xaml.cs
+++ b/Content.Client/_DV/Traits/UI/TraitCategory.xaml.cs
@@ -183,6 +183,16 @@ public sealed partial class TraitCategory : BoxContainer
         UpdateStats();
     }
 
+    //Euphoria | Clears previously set conditions.
+    public void ResetConditions()
+    {
+        foreach (var (_, entry) in _traitEntries)
+        {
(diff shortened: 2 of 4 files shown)


Cprn-Sprkl:
This takes care of part of the issue, but not all of it. Will have to cook it a bit more when there's more time.

Mnemotechnician (approved):

Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
```diff
@@ -588,8 +588,8 @@ private void UpdateTraitsSelection()
                 }
             }
 
-            Traits.SetSelectedTraits(selectedTraits);
             Traits.UpdateConditions(Profile);
+            Traits.SetSelectedTraits(selectedTraits);
```
Mnemotechnician: It seems that after your change, the game is going to first update conditions (on the OLD profile) and only then change the profile. This doesn't make sense to me personally.

Cprn-Sprkl: I'm not sure what you mean. This method gets called in a few situations, including when you change character slots. UpdateConditions being at the bottom of the method was causing an issue where the traits from the new profile were trying to apply while the conditions hadn't been updated for the new profile. Swapping these two causes the trait tab to work as expected.

Here is a step by step debug where I swapped from character slot 1 to character slot 0. You can see the profile it is checking for UpdateConditions is the slot 0 one:
image


The reason this is currently a draft is because while the trait tab is visually working as expected, the profile itself is remaining dirty and isn't actually applying the trait as expected behind the scenes.

image

image

image

Note that in the last picture the save button is available. If you attempt to leave character select it prompts you to save, if you do and then come back in you'll find that it now shows this:
image

It is entirely possible I am misinterpreting the debugger results and code though.

Mnemotechnician: The logic is "change current traits to match the new profile" -> "validate new traits". You changed the logic to "validate OLD traits" -> "change currrent traits", which doesn't make sense. You'd achieve the same effect by commenting the line out.

Cprn-Sprkl: Commenting the line out does not give the same result. It gets rid of trait locks all together:
image
image

I was concentrated on how it was working through the UI and not considering that is also served as a validator. The commented out behaviour did give me an idea though and it seems to have worked. I just need to add Euphoria tags to everything and run a final test before I commit. It's looking promising though.

Cprn-Sprkl:
I could have maybe modified the existing UpdateCondition stuff to handle resetting them when passed a bool, but I figured this would be a bit cleaner in case of upstream changes, despite a bit of code duplicating.

Mnemotechnician (approved):
Alright, this seems like a better solution.
For context, trait validation is needed so that e.g. if a trait prototype gets removed, the player's profile will be updated to have the now-invalid trait removed (otherwise there's no way for a player to get rid of a trait with a missing prototype without exporting and manually editing your character's profile)

PR #611, opened 2026-05-23
Extend lobby duration

Oshmoo:
## About the PR
Extends the lobby's duration to 5 minutes up from 2 minutes and 30 seconds. Also extends the automated lobby/gamemode votes to compensate.

## Why / Balance

A byproduct of the discussion in our discord feature request thread "Make half of the current unreadied players count towards secret gamemode rolls" made a few people bring up that the lobby timer seems to be too short for them for one reason or another. This lets people wind down for a little bit longer in-between rounds without making them wait too long and hopefully allow a few more people to ready up at the start of the round.

## Technical details

First of my many victories on the YAML battlegrounds

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- tweak: The lobby duration has been extended to 5 minutes, up from 2 minutes and 30 seconds

### DIFF
--- a/Content.Shared/CCVar/CCVars.Game.cs
+++ b/Content.Shared/CCVar/CCVars.Game.cs
@@ -21,7 +21,7 @@ public sealed partial class CCVars
     ///     Controls the duration of the lobby timer in seconds. Defaults to 2 minutes and 30 seconds.
     /// </summary>
     public static readonly CVarDef<int>
-        GameLobbyDuration = CVarDef.Create("game.lobbyduration", 150, CVar.ARCHIVE);
+        GameLobbyDuration = CVarDef.Create("game.lobbyduration", 300, CVar.ARCHIVE); // euphoria - let people wind down between the 2-3+ hour rounds we tend to have
 
     /// <summary>
     ///     Controls if players can latejoin at all.
--- a/Content.Shared/CCVar/CCVars.Vote.cs
+++ b/Content.Shared/CCVar/CCVars.Vote.cs
@@ -62,7 +62,7 @@ public sealed partial class CCVars
     ///     Sets the duration of the map vote timer.
     /// </summary>
     public static readonly CVarDef<int>
-        VoteTimerMap = CVarDef.Create("vote.timermap", 90, CVar.SERVERONLY);
+        VoteTimerMap = CVarDef.Create("vote.timermap", 180, CVar.SERVERONLY); // euphoria - compensate for lobby duration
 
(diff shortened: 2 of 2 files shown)


kivdon:
This seems like a good candidate for polling the entire Discord first. This change affects literally every single person, and some people have already voiced that they don't want it.

Barheru:
Personally, I support this change. Fixing the low antag count that has come from so few people being readied up is important--and it could be tricky to find a real solution to that programmatically. If this change can be affected socially, that would save our contributors and maintainers a lot of time. 

This can also easily be changed back if the community hates it, but personally, I think waiting 2 minutes and 30 seconds extra between rounds is fine. It lets us refill our water, make that loadout change, etc etc, without worrying that we won't get the coveted chaplain slot because we were playing around in the character creator.

kivdon:
For anyone who doesn't want to dive through the thread, here's a summary of some of the reasons people have given for being against extending the lobby time:

1: Not everyone needs extra time, but this would affect everyone regardless. There have been no polls, so it's impossible to say how many people do or don't need extra time. But some don't like the idea of everyone being forced to wait longer just because some people need more time. 

2: This would also extend the already lengthy time between evac being called and a new round starting. A lot of people don't want to join the round if it's about to end, so that means they have to wait until the next one starts. That's already about 20 minutes after evac gets called. Making that even longer is not pleasant, especially for those who have very limited time. 

3: Probably about half (maybe a bit more) of the people who've given a reason for not readying up, have had that reason be unrelated to needing more time. Many people have stated that they late-join on purpose, so they can pick roles based on what's taken and who's where. No matter how long the lobby timer is, a whole lot of people are still going to late-join anyway. 

If the reason for wanting to extend the lobby timer is so that you have more time before the round starts to get up and do something, then either hit "Ready" and start playing when you're back at the keyboard, or simply late-join. 
If the reason is to encourage more people to ready-up to get more antags, this is barely going to help, since a lot of people late-join on purpose. This is a problem that should be solved through other means, such as making unreadied people count towards antag rolls in some way. Not by forcing _everyone_ to wait longer.

Mnemotechnician (approved):
This is NOT the correct way to do cvar changes, but this'll do.

SyaoranFox:
DNM as we test the changes to player calculation

SyaoranFox:
Due to the player calculation changes, I believe this is no longer required. Thank you for taking the time to try to fix the problem at hand.

PR #615, opened 2026-05-24
Salvage magnet

Raikyr0:
## About the PR
Changes the cost of wrecks pulls from 500 to 50.

## Why
Shits too expensive for wrecks that can just give fucking nothing.

**Changelog**
:cl:
- tweak: Changed Salvage Magnet cost from 500 to 50, shit was not worth 500 points.


SyaoranFox (approved):
We can change this later down the track if its too low, or we add better wrecks.

Mnemotechnician:
I'm not sure a 10x change is justified here

Raikyr0:
> I'm not sure a 10x change is justified here

The issue is, the random wrecks which cost 500 mining points typically, are less valuable then the randomly generated free ones in terms of materials and loot, and more dangerous, but I worry that if I set it to free the system DV has for it would break.

SyaoranFox:
Due to lacking a second approval, I am closing this PR. While I do agree that the price for wrecks is quite high, it is egregiously high. Salvage does end up with lots of excess points. This problem may be tackled by increasing the chances of better loot rather than reducing the cost to get substandard loot.

PR #635, opened 2026-05-27
Content Warning: Vore Patch 3

JaDiDo:
## About the PR

1. Rework of the VoreSystem consent
2. Better handling of components 
3. Moving prey related methods into a new separate file (DevourSystem)

## Why / Balance

1. Enforcing stricter and more clearer rules for consent, entity check like dealing with possible scenarios where the prey has the be ejected to avoid further consent breaches + clearer structure through methods
2. Restructure of how the system deals with components in a separate file for maintenance for later creation of client systems such as customized blindness in future patches 
3. To debloat VoreSystem as it previously handled both pred+prey related issues and doesnt look suitable for longterm maintenance

## Technical details

Handling of immunities were moved to the new DevourSystem file and VoreImmunityComponent renamed to DevourComponent to better showcase its usage. Furthermore new methods/functions:
1. On dead/crit prey eject: OnPreyMobStateChanged
- Upon changes of the mob state of a prey inside it will release the specific prey from ALL related vore_container (in case of multi vore hence the loop (includes a safetynet)). It handles all the regular issues regarding immunity components and cords
- To avoid rot and alike avoid med consent breach (given that this patch also deals with cordmanagement)

2. Helper method for verbs: IsDevourable
- Instead of using the GetVerbs system for cheks IsDevourable will now do it to see if its a valid interaction and containment + has some new altered checks to be more strict in regards of consent and gameplay
- Certain scenarios were allowed such as admin ghost vore for example (only BodyComponent check for one person previoulsy). Now it should be prevented altogether. Also included: avoiding verbs showing up for crit/dead to prevent rot/accidentally RR, checks for playersession instead of MindComponent meaning no interaction possible when one person leaves the game for example. BodyComponent removed during consent check to be assigned for everyone with consent on in order for it later to be used for client related code such as later making a menu for toggles/customization (since its now in this new method)

3. Dealing with sensors when giving Immunities
- Instead of applying it to digestion only sensors will be now turned off/on when entering/leaving a vorecontainer when handing out immunities
- This way should be unlikely for med or sec to stumble on a scene with a prey inside when trying to search them using a console monitor and due how the system now ejects crit and dead + making those inedible there should be no risk of accidental RR, worked for both for release and escapes. It also supressed/reactivated medical implants

4. Update check for component during consent check
- A safe check inside the Update method 
- During testing it showed to be unsafe to let it run without a component check in case for example of entity deletions hence added a component check that will otherwise just clear the values. The update for immunity did not showed any issues hence it remained the same

5. Comp call in most methods removed in VoreSystem and moved to IsInVoreContainer
- Most of the time VoreComponent was only needed for a IsInVoreContainer check
- Cleaner code structure, less reliance for methods

6. Rework of escape (deletion of VoreComponents boolean and turning off escape by movement)
- A new verb for devoured people when self inspecting to be able to leave the container instead of the previous movementbased escape
- In previous System it was too easy to accidentally attempt to escape when simply typing + the used boolean for meant escapes made the code too confusing through its reference all over the code

## Media

Screenshot 2026-05-27 173641

</p></details>

## Requirements

## Licensing:

**Changelog**

:cl:
- tweak: Rework of the VoreSystems consent


JaDiDo:
MergeConflicts should be resolved

Content.Server/_Floof/Vore/DevouredSystem.cs
```diff
@@ -0,0 +1,205 @@
...
+        }
+        if (tracker.AddedTemperature){
+            RemComp<TemperatureImmunityComponent>(prey);
+            tracker.AddedTemperature = false;
+        }
+        if (tracker.AddedRadiation){
+            RemComp<RadiationProtectionComponent>(prey);
+            tracker.AddedRadiation = false;
+        }
+        _suitSensorSystem.SetAllSensors(prey, SuitSensorMode.SensorCords);
```
Mnemotechnician, line 202: You're always enabling suit sensors regardless of what their previous value was. That's suboptimal.

Mnemotechnician (approved):
Looks fine besides that

Content.Server/_Floof/Vore/DevouredSystem.cs
JaDiDo: Thank you for checking it out! Does make sense will make sure to patch it out

PR #645, opened 2026-05-30
Rrrolling Accent

Cprn-Sprkl:
## About the PR

An implementation of the Tajaran accent from the Den (added there by chuga-git) as a non-species specific accent trait. It's now a selectable trait called Rrrolling Accent. The system was rewritten, but the concept is the same.

## Why / Balance

The accent was requested.

## Technical details

N/A

## Media

image

</p></details>

## Requirements

## Licensing:

## Breaking changes

N/A

**Changelog**

:cl: sprkl
- add: Added an accent for rolling your 'r's.


Mnemotechnician (commented):
This is... so buggy and inefficient. They are splitting the message into words and performing random regex replacement in each word. So if you have a world like "roar", it it will become either "roar", "rroarr", or "rrroarrr", but not something in between. This could be achieved with just 1 regex replacement (replace `(r|R){1,3}` with `$1$1$1`):

Image

Cprn-Sprkl:
Was bringing it mostly as is for the people that requested it, but I'll take a look at actually improving it after work. Make 'roar' become 'rroarrr' sometimes. Probably also tackle the issue where 'Roar' would be 'RRRoarrr'. It's not the most pleasant read.

Mnemotechnician (approved):
Yeah, this is much better.

M3739 (approved):

PR #666, opened 2026-06-05
Conversion immune trait

Mnemotechnician:
## About the PR
Lets you opt out of conversion antags. Currently only covers the cosmic cult but can be extended to cover other antags as well.

Relevant antags and aghosts will see additional examine an an icon above the player with the trait:
image
image

Old demonstration:

https://github.com/user-attachments/assets/267f453d-1cd4-4c29-a051-bdf31e631285

## Licensing:

## Breaking changes
My sleep schedul

**Changelog**X
:cl:
- add: Players can now choose to opt out from Cosmic Cult conversions via traits.


Resources/Locale/en-US/_Floof/traits/traits.ftl
```diff
@@ -22,6 +22,12 @@ trait-marked-remove-desc = Someone or something wants you REMOVED. Antags can re
 trait-marked-teach-name = Marked for Death
 trait-marked-teach-desc = Someone or something wants you taught a lesson. Antags can recieve objectives to kill you.
 
+trait-conversion-immune-name = Conversion Immune
+trait-conversion-immune-desc =
+    This trait prevents you from being converted mid-round by certain antags.
+    Currently only affects the Coscmic Cult.
```
M3739: Minor proofread
```suggestion
    Currently only affects the Cosmic Cult.
```

Resources/Locale/en-US/_Floof/traits/traits.ftl
```diff
@@ -22,6 +22,12 @@ trait-marked-remove-desc = Someone or something wants you REMOVED. Antags can re
 trait-marked-teach-name = Marked for Death
 trait-marked-teach-desc = Someone or something wants you taught a lesson. Antags can recieve objectives to kill you.
 
+trait-conversion-immune-name = Conversion Immune
```
M3739: 'Immune' in the name sounds more like a defense/performance relation. Changed to more neutral wording that remains accurate.
```suggestion
trait-conversion-immune-name = Unconvertible
```

Resources/Locale/en-US/_Floof/traits/traits.ftl
```diff
@@ -22,6 +22,12 @@ trait-marked-remove-desc = Someone or something wants you REMOVED. Antags can re
...
+trait-conversion-immune-desc =
+    This trait prevents you from being converted mid-round by certain antags.
+    Currently only affects the Coscmic Cult.
+    Keep in mind that this does NOT in any way protect you from getting killed in case you try to use the immunity to your advantage.
```
M3739: Rewrite for a bit more brevity and clarity
```suggestion
    NOTE: This trait DOES NOT protect you from being killed if you abuse it (e.g baiting antags into failing to convert you).
```

Resources/Prototypes/_Floof/Traits/meta.yml
```diff
@@ -35,3 +37,20 @@
...
       targetType: Remove
+
+- type: trait
+  id: ConversionImmune
+  name: trait-conversion-immune-name
+  description: trait-conversion-immune-desc
+  category: Meta
+  cost: 1
+  usesSlots: false
+  conditions:
```
M3739, line 48: No self-defeating immunity.
```suggestion
  conditions:
  - !type:IsAntagEligibleCondition
    antag: CosmicAntagCultist # Can you sell faith you can't lose?
    invert: true
```

M3739 (changes requested):
After much thought on my end, I have an issue with the PR as it currently stands.

First of all, the implementation looks sound and clearly works. The issue I have lies with presentation and framing.

I... don't feel comfortable attaching a mark on people because they don't *want* to engage in certain things. In social contexts, I would ideally like to avoid, from the antag's perspective, the *negative* side of this system being the only feedback they get. It'll result in more problematic conversion antags as a result, as well as warping public perception on the system. Negative feedback will foster negative effects.

So, **I propose we reframe it, and instead make it mark those who *can* be converted.** This way, the antag is getting *positive* feedback in the form of this person being 'For Hire' rather than 'Don't recruit me'. There is a difference between 'I couldn't find potential targets' versus 'nobody here is *permitting* me to convert them.'

<ins>TL;DR - Change the presentation to reframe it as positive feedback, instead of negative feedback to conversion antags, by instead marking who *can* be converted rather than who *doesn't want to be* converted.</ins>

Mnemotechnician:
> I propose we reframe it, and instead make it mark those who can be converted

That is going to cause a lot of visual noise. Personally I'm already not happy with the heads-up icon being there in the first place, adding even more clutter is only going to make it worse. It is **meant** to give negative feedback - the minority who opts out of the cult is not to be interacted with unless necessary.

M3739:
> > I propose we reframe it, and instead make it mark those who can be converted
> 
> That is going to cause a lot of visual noise. Personally I'm already not happy with the heads-up icon being there in the first place, adding even more clutter is only going to make it worse. It is **meant** to give negative feedback - the minority who opts out of the cult is not to be interacted with unless necessary.

I see.

You are correct on that regard, and I stand corrected. The visual clutter can be hellish. And I do share your sentiment on the HUD icon being there to begin with.

Perhaps I am being a bit too paranoid. I suppose we shall keep the original design language as is.

Then from it's results, we shall adjust as neccessary.

M3739 (approved):

SyaoranFox:
I'll merge this PR when Admins have their rules/enforcement policy in place.

PR #671, opened 2026-06-07
Organ Changes

Raikyr0:
## About the PR
This PR homogenizes the amount of chemicals players can fit into their stomachs to 50 units.
Also changes "mouse" stomach to "critter" stomach because it was also used for snails, and decreased the consumption amount to 10 as well as added a mothroach variant with also only 10.
This PR also gives Rodentia a 75 unit stomach to compensate for their double hunger rate.
Also, harpies can eat chocolate again, because they're more human then animal.

I don't know how much organ work I want to actually do when I know Newbody will break all of it again.

## Why / Balance
Uniforms the amount people can eat, with the exception of Rodentia who has an increased amount due to hunger rate.
Changes to critter and mothroach stomachs prevents them from consuming their whole body weight in 1 go and emptying the kitchen.

**Changelog**
:cl:
- add: Added Critter, Mothroach and Rodentia Stomachs. Critter and Mothroach stomachs have a much lower consumption limit which prevents them from hoovering up the whole kitchen spread, while Rodentia stomachs increases how much they can shove inside.
- tweak: Changed most races to be able to consume 50 units at once.
- tweak: Changed Harpies to be able to eat chocolate and tea.


SyaoranFox:
This is a lot of changes that feel unrelated to one another. Mice and mothroaches are meant to be antags towards leaving items around the station. They also clean up unwanted entities by consuming them. The harpie change I can get behind, but birds shouldn't be eating chocolate/tea, but they are also like only half bird, so its a tough one to discern. The downside to rodentia is that they NEED to eat more. Giving them a larger stomach feels like it counters the need to eat more.

I am not sure these changes are needed.

Raikyr0:
> This is a lot of changes that feel unrelated to one another. Mice and mothroaches are meant to be antags towards leaving items around the station. They also clean up unwanted entities by consuming them. The harpie change I can get behind, but birds shouldn't be eating chocolate/tea, but they are also like only half bird, so its a tough one to discern. The downside to rodentia is that they NEED to eat more. Giving them a larger stomach feels like it counters the need to eat more.
> 
> I am not sure these changes are needed.

Giving Rodentia a bigger stomach allows them to swallow more food on the table at once. Once they eat food (like a whole pizza which normal people can eat in 1 go), they'll hit the max hunger cap regardless.

Harpy wise, they're supposed to be genemodded humans, so I believe they would be fine. (Felinids require it to eat mice tho.)

As for the pest, yes they're supposed to eat unattended food like a food antag, but it didn't feel too realistic a mouse could devour 40u of food in 1 go. So what will end up happening with the change is they'll take two bites, metabolize it and then continue, assuming the mouse is still hungry, which leaves the food on the table still but it's now 2 bites less. (like finding out your food is hollow)

PR #677, opened 2026-06-08
Anomaly Hats

Dunrab:
## About the PR
Ever thought well damn this infection anom affect is really cool, wish i could keep it with out risking dying to an anom? Well now you can.

## Why / Balance
Something I thought would be a cool addition so figured id make it possible. I let the hats be put on pets like hampsters and moth roaches, they don't display perfectly but i like the idea of hammy stealing an anom core to have a hat.

## Technical details
Probably could have done the construction graphs in a way better way than i did, some yaml, tag and ftl edits. I gave tags to each inert anomaly core since they only had one tag they all shared

## Media

image

image

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- add: Inert anomaly cores can now be turned into hats. These hats will have the visual affects of infection anoms.

### DIFF
--- a/Resources/Locale/en-US/_Floof/recipes/tags.ftl
+++ b/Resources/Locale/en-US/_Floof/recipes/tags.ftl
@@ -20,3 +20,15 @@ construction-graph-tag-carpet-simple-white = carpet simple white
 
 #masks
 construction-graph-tag-welding-mask = welding mask
+
+#hats
+construction-graph-tag-bluespaceanomcore = anomaly core
+construction-graph-tag-floralanaomlycore = anomaly core
+construction-graph-tag-firealanaomlycore = anomaly core
+construction-graph-tag-fleshalanaomlycore = anomaly core
+construction-graph-tag-frostalanaomlycore = anomaly core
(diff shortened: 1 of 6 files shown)


Dunrab:
I feel like I talked to someone about this in the past, can't remember who but here you are if you see this.

M3739:
I... genuinely don't know how to feel about this.

On one hand, it looks cool, I agree. On another hand we're basically gaslighting Epistemics, which while really funny during a code white, has it's own implications.

Dunrab:
> I... genuinely don't know how to feel about this.
> 
> On one hand, it looks cool, I agree. On another hand we're basically gaslighting Epistemics, which while really funny during a code white, has it's own implications.

Epi has the tools in t1 research to find and locate anomalys, if they aren't using it then well thats on them. Epi players should also be taking cores for a few reason after they kill an anom if they aren't then well thats another on them. I look at it like loto oil, epi should be controlling who has access to it like who gets access to these hats.

M3739 (approved):
That is a fair argument to make.

Besides, with Lustro Ecclesia existing with their fire halo, I suppose it'll work out.

PR #685, opened 2026-06-10
Added Shadekin Light Damage

Raikyr0:
## About the PR
Adds Light damage to Shadekin, they now take heat damage in the light and slowly heal in the dark. With a 10% slow in light and 10% speed boost in darkness.

:) I left heal when dead on so medical needs to fix them in a dark room.

## Why / Balance
Lore reasons.

## Technical details
Uses pre-existing components from Diona and Skia.

**Changelog**
:cl:
- add: Shadekin are now properly allergic to light again.

### DIFF
--- a/Resources/Prototypes/_Floof/Entities/Mobs/Species/shadowkin.yml
+++ b/Resources/Prototypes/_Floof/Entities/Mobs/Species/shadowkin.yml
@@ -147,7 +147,7 @@
         Toxin: -0.15
         Airloss: -0.15
     lightDamage:
-      groups:
+      types:
         Heat: 1
     healWhenDead: true
     darkMovementSpeedMultiplier: 1.1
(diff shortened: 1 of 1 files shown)


kivdon:
I could be wrong, but don't Shadekin have no benefits currently? I know they have lower Crit and Dead thresholds than most. 
Though I also don't remember what benefits they used to have in Floof.

Arbitrated:
Shadekin don't breathe, which is normally a benefit but occasionally causes problems if they somehow take Asphyxiation damage. 

The problem with this feature is that it's going to make it very difficult for a Shadekin to, well, [i]exist[/i] on a station, unless they're playing as a Maints Goblin and never step out into the public. 

Do we know what lore is being referenced that indicates Shadekin are actively harmed, possibly even killed, by bright lights?

Raikyr0:
It used to provide a detriment to mood, which directly affected their speed.

It is currently just set to the reverse of Diona in terms of damage taken and healed. (numbers adjustable)
I considered this because it would create additional roleplay when dealing with Shadowkin.

Mnemotechnician:
Either make a server-wide poll to see if people would like this change or make it optional

SyaoranFox:
I am gonna say no to this because it is adding a detriment with no real benefit to speak of. Shadekin are in a weird spot right now, and are in the process of being reviewed. Once that is done (no, I am not sure when that will be), you may re-PR this.

PR #698, opened 2026-06-13
Autocall Secret Ballot And Tiebreaker Fax

M3739:
## About the PR

This PR aims to change the automatic evacuation shuttle call (autocall) to do two things instead:
- No longer show votes to players, while still showing votes to admins.
- If a tie occurs, sends a fax to the Captain's Office with instructions.

Instead of porting over the original secret ballot system implemented pre-rebase, I elected to rebuild and redesign the system from scratch in order to better fit it's intended function, while retaining the OOC nature of the original voting system itself.

I will admit, I feel *shoddy* as a programmer, but that could be my imposter syndrome talking. Much of what you see here is cobbled together from examples and observations I made, glued together with my tech-agnostic intuition. So if you do notice any issues, or any possible way I can do this in a better manner, please do so. I am more than willing to learn in that regard.

## Why / Balance

Currently, we have been noticing an uptick in the community in regards to tying the vote intentionally, or shifting the vote at the last minute. Not only does this frustrate the community, and cause rounds to end earlier than intended, but it makes it harder for both admins and eventmasters to read the room, so to speak.

In addition, this will greatly curtail the bandwagon effect that dilutes the truthfulness of the vote. With this system, we can expect more truthful and honest voting moving forward, while also curtailing the issues with our pre-rebase implementation due to it being IC (e.g 'I voted stay because my character would vote stay').

Oh yeah, and in case if this isn't obvious, this is feature parity with pre-rebase.

## Technical details

image

In order to pull this off, we swapped out DV's method, `CallEvacuationVote`, with our own variant of it, named `CallEvacuationSecretBallot`. The structure is similar, but with two key differences:

- Vote is created with `DisplayVotes = false`, which hides the live count of players voting for either option. Admins will still see the live count regardless. As a matter of fact, the summary for the parameter states that for Admins, it will always be `true` client-side.
- if `args.Winner == null`, it will call evac, but also call our second new method, `SendShiftExtensionReview` and set it's returned value as a variable. Said variable is then used to determine whether or not said new method succeeded or failed. On fail, it will notify admins of such.

`SendShiftExtensionReview` will first attempt to enumerate all known fax machines, then flip through each entry looking for a fax machine containing the field `ReceiveShiftExtensionReview`. This is similar to how sending nuke codes to the station resolves which fax machine to send it to. When a match is found, it will construct the printout. Said printout is locked to prevent tampering IC as it bears a CC stamp. Then it sends the fax, and sets `wasSent`, the return boolean, to true.

If it did send a fax, at least one fax, it will send a global announcement with the relevant context to alert command to read it.

Note that due to the fluent parser being not so fluent nor clear with it's errors, I had to leverage an *insane* work around seen in our other documents that leverage fluent for their body: Using a Braille ASCII character: 0 dots, which appears as an empty space (⠀). Virtually indistinguishable from a normal space unless you have the option to show whitespace enabled.

## Media

image

image

image

image

https://github.com/user-attachments/assets/e28dcca0-bdac-4fc5-957c-7c31fa46071a

</p></details>

## Requirements

## Licensing:

## Breaking changes

Not... that... I... know of?... I hope.

**Changelog**

:cl:
- tweak: The automatic evacuation shuttle call is now a secret ballot; live counts of each option is no longer visible for non-admins.
- add: The automatic evacuation shuttle call now has an IC tiebreaker system.


Arbitrated:
Awesomesauce! <3 

What does the announcement look like for when a vote is tied and the SER form successfully sends? We could see the "tie" message being easily missed, especially as we've seen several Command members fail to notice their name being called out on Command Comms...

We were going to express some concern at a Captain or ACO not being present but uh. then we remembered that the Cap ID card """unlocks""" automatically anyways, which means if an ACO or Captain doesn't exist, another commmand member can just take the card and enter Cap's room to check the fax anyways.

Content.Shared/Fax/Components/FaxMachineComponent.cs
```diff
@@ -142,7 +142,15 @@ public sealed partial class FaxMachineComponent : Component
...
     [DataField]
     public bool ReceiveStationGoal { get; set; } = false;
+
     /// Floofstation addition end
+    /// Euphoria addition start
+    /// <summary>
+    ///     Should this fax receive shift extension reviews?
+    /// </summary>
+    [DataField]
+    public bool ReceiveShiftExtensionReview { get; set; } = false;
```
Mnemotechnician: This seems redundant as this will always be limited to the captain's office

Mnemotechnician (approved):

SyaoranFox (approved):

Content.Shared/Fax/Components/FaxMachineComponent.cs
M3739: I would beg to differ. This is to establish the field that will be utilized as a flag. I followed suit similar to how the nuke codes and station goal find the suitable fax machines (See `ReceiveNukeCodes` and `ReceiveStationGoal` and how they are utilized). Without this, the code right now as it stands will grind to a halt.

In addition, it allows for other printers to be marked for receiving said documents should it be needed, so expandability on that regard. It's just that as of writing, the only unique fax machine prototype is indeed the Captain's fax.

So, in actuality, this field is what makes it *limited* to the captain's office in the first place.

PR #700, opened 2026-06-14
Arrivals Terminal Updates

Dunrab:
## About the PR
Small updates to the arrivals terminal.

## Why / Balance
Needed to remove some gamer loot and fix up some areas.

## Technical details
Removed the cryosleep pods in order to prevent a manifest bug, if we ever happen to fix that bug then we can add them back. It uses late join spawners in its place which will work the same.

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- tweak: Some small updates have been made to the Arrivals Terminal.


thefrostbite11037:
Are you saying you're removing all of the cryosleep pods from arrivals? In that case, could you be more specific on what the manifest bug is? Because I rely on those pods to deal with a traits bug that I get every so often that apparently no one has ever fixed, and having no ability to cryo at arrivals WILL screw me over pretty badly.

SyaoranFox (approved):

PR #708, opened 2026-06-15
Fix order-dependant trait validation

Eagle0600:
## About the PR
Fixed by ordering all free and negative-point cost traits first, then most expensive to least expensive traits (this means lower-cost traits are lost first if too many points are spent).

## Why / Balance
Inconsistent trait resolution order means that some traits were rejected due to being too expensive, even when the final point total would have been correct.

## Technical details
The ValidateTraits method of TraitSystem sorts traits into two regions before running individual trait validation:
1. Traits with a cost <= 0, in descending order.
2. Traits with a cost > 0, in descending order.

Order of traits in the region 1 is unimportant, with a matching sort order to region 2 for simplicity of implementation.
## Requirements

## Licensing:

## Breaking changes
Changes to the ValidateTraits method should be non-breaking (input and output parameters are unchanged).

**Changelog**
:cl:
- fix: Traits should no longer be erroneously rejected for being too expensive.

### DIFF
--- a/Content.Server/_DV/Traits/TraitSystem.cs
+++ b/Content.Server/_DV/Traits/TraitSystem.cs
@@ -185,20 +185,8 @@ public sealed class TraitSystem : EntitySystem
         }
 
         // Sort negative and zero point costs first, and then lower costs last. The order of the negative and zero-cost traits is unimportant.
-        resolvedTraits.Sort((a, b) => {
-            if (a.Cost > 0 != b.Cost > 0)
-            {
-                // A and B have different signs. Sort in ascending order.
-                return a.Cost.CompareTo(b.Cost);
-            }
-            else
-            {
-                // A and B have the same sign. Sort in descending order.
-                return b.Cost.CompareTo(a.Cost);
-            }
-        });
-
-        foreach (var trait in resolvedTraits) {
+        foreach (var trait in resolvedTraits.OrderBy((trait) => trait.Cost > 0).ThenByDescending((trait) => trait.Cost))
+        {
             var traitId = trait.ID;
 
             #endregion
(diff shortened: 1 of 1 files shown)


Mnemotechnician (changes requested):
Use OrderBy and ThenBy, see line 120. Ideally, if you can, make a new method to perform the sorting so that we don't have to change it in two places the next time.

Eagle0600:
> Use OrderBy and ThenBy, see line 120. Ideally, if you can, make a new method to perform the sorting so that we don't have to change it in two places the next time.

To be clear, are you suggesting a static method on the TraitSystem for sorting a collection of traitprototypes? Would you want it to be something like SortTraitsPrototypesForValidation?

Mnemotechnician:
Whatever works, there are at least two places where this kind of sorting needs to be done. It needs to be consistent, otherwise trait validation and trait application may diverge.

Eagle0600:
> Whatever works, there are at least two places where this kind of sorting needs to be done. It needs to be consistent, otherwise trait validation and trait application may diverge.

The second place it's sorted is for application. Applying performs no checks and validating performs no application. None of the same work is done. There should be no need to sort before applying at this point, and I'm actually just going to remove that.

edit:
To give a rational for removing the two sorts before applying traits other than removing redundant code: It makes OnPlayerSpawnComplete and OnGhostRoleSpawnerUsed the same as the DeltaV version, improving code maintainability if they ever make changes. I am still moving sorting to a static method.

Mnemotechnician:
> There should be no need to sort before applying at this point, and I'm actually just going to remove that.

During trait application, traits are ordered by priority first, and priority DOES matter. Though ordering by cost MAY be redundant there - i'm not sure and don't want to dig into the code right now.

Eagle0600:
> > There should be no need to sort before applying at this point, and I'm actually just going to remove that.
> 
> During trait application, traits are ordered by priority first, and priority DOES matter. Though ordering by cost MAY be redundant there - i'm not sure and don't want to dig into the code right now.

After shooting my mouth and actually digging into the changes involved, I've come to the same conclusion. We want them ordered by priority for application and cost for validation, two different use cases. Consequently, I'm creating two different static methods.

SyaoranFox (approved):

Mnemotechnician (approved):

PR #712, opened 2026-06-16
Content Warning: VoreSystem new content 1

JaDiDo:
## About the PR

New content for the VoreSystem to improve flavour for container transfer and the containment itself.

## Why / Balance

Current system had a lack of feedback from being transferred into the vore container (1) and for being inside said container (2). As the system grows we also need dedicated categories for the verbs (3) to remove the cluster for example: removing individually prey instead of all at once. The new verb (4) will allow more interaction with the system that should allow new possible scenarios.
I have asked the vore community with a poll in the discord vore channel. 10 participants have agreed to all changes with no rejections. 

## Technical details

1. Gulp sound effect
- playEntity will only play it for the 2 involved entities, the sound used is the same as the previous version 
- A requested change but previous system had reports of people hearing through walls, problem was Filter.Pvs(uid) was used instead of directly using related sessions (see floofsystem commit 493efb1). It got eventually patched in the mentioned commit with playentity, I tested it with several clients and it works without issues.

2. Client based vision change
- once entity gains the DevouredComponent they will have a stomachlike circle with a dimming of the screen
- Before there was a lack of visuals from being inside the stomach. Now a circlemovement through the cosine will simulate being inside a stomach_container which in later patches will be made more customizable. I tested it for different overlays such as oxy/brute damage and blind or alcohol poisoning. Overlayed but still visually working. Also works when client relogs through the overlay system. 

3. VerbCategories
- Verbs will be assigned to categories which contents are stored in a separate file
- Reached a point where too many verbs for the system are there. With it we should have a cleaner menu + making it harder to initiate it by accident. DigestCategory is not assigned anywhere yet but will be used later+showcases how I am planning to separate the different verbs (will later include icons once created)

4. Insert someone else
- A new verb interaction for when you either carry or drag someone to be inserted (similar to the other current working verbs devour and insert yourself but with the entity carried/pulled being transferred as prey to target)
- After asking was requested by the community. Only people who have VoreComponent (therefor have either pred or prey enabled) see the verb which I think should be fine in terms of consent since all participants already agreed to be exposed to vore related elements in the first place. Through the method isDevourable it should resolve every other issues regarding consent hence prey toggle/not offline and such

Also moved CVars into a subfolder alongside Verbs for better separation and renamed VoreImmunitySystem to DevouredSystem for better clarification

## Media

https://github.com/user-attachments/assets/8fc18d7f-44fd-48a5-831a-08b97d14ba53

Screenshot 2026-06-16 164325

</p></details>

## Requirements

## Licensing:

**Changelog**

:cl:
- add: improved flavour for VoreSystem regarding containment and container transfer


Content.Client/_Floof/Vore/DevouredBlindSystem.cs
```diff
@@ -0,0 +1,47 @@
...
+            if (_player.LocalEntity == uid)
+                _overlayMan.AddOverlay(_overlay);
+        }
+
+        private void OnVoreShutdown(EntityUid uid, DevouredComponent component, ComponentShutdown args)
+        {
+            if (_player.LocalEntity == uid)
+                _overlayMan.RemoveOverlay(_overlay);
+        }
+    }
```
Mnemotechnician, line 47: This is overindented. Not a big deal.

Content.Shared/_Floof/Vore/Other/VoreVerbCategory.cs
```diff
@@ -0,0 +1,12 @@
...
+namespace Content.Shared._Floof.Vore;
+
+public static class VoreVerbCategory
+{
+    public static readonly VerbCategory VoreGeneral =
+        new("Vore", null);
+
+    public static readonly VerbCategory VoreDigest =
+        new("Digest", null);
+}
```
Mnemotechnician, line 12: I'd put these category definitions inside the systems that add them.

Mnemotechnician (approved):
Looks more or less fine code-wise.

PR #721, opened 2026-06-19
Vector to use 45 auto

Raikyr0:
## About the PR
This PR fixes the Security Vector by giving it 45 auto instead of magnum.
This PR decreases the Vector's damage by 33%, but decreases it's spread. 
(I rolled back on the burst and mag size changes because those are based on it's real sizes)

Codewise - Creates a new ammo type (45 Auto) which uses the (35 auto sprites) and replaces the Vector's SMG mags to use that. 

Future stuff - Maybe make the Universal also shoot 45 auto? I don't mind the admin 1984 pistol still shooting magnum tho.

## Why / Balance
Why? They made a gun that shoots 45 ACP shoot 45 magnum.
Balance, 45 auto projectile damage is in the middle (actual half way point rounded up for odd numbers) between 35 auto and 45 magnum. It's still very good, but also the gun is very expensive still. 

**Changelog**
:cl:
- tweak: Tweaked Vector (Security) to fire 45 auto, which decreases it's damage output by 33%, decreased it's spread slightly.

### DIFF
--- a/Resources/Prototypes/Recipes/Lathes/Packs/security.yml
+++ b/Resources/Prototypes/Recipes/Lathes/Packs/security.yml
@@ -32,6 +32,7 @@
   - SpeedLoaderMagnumPractice
   - MagazinePistolSubMachineGunPractice
   - WeaponFlareGunSecurity
+  - MagazineBox45PistolPractice # Euphoria
 
 # Shared between secfab and emagged autolathe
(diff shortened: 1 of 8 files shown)


Raikyr0:
Closed to convert it to use 35 Auto after a short chat with Tech about it.

PR #722, opened 2026-06-19
Vector

Raikyr0:
## About the PR
This PR changes the Vector to shoot 35 auto (at the suggestion of Tech because making a 45 ACP ammo type is more work then just swapping it and the gun sprite is the 9x19mm one). 

## Why / Balance
Why, because it shot 45 magnum before.
Balance, it's still expensive as hell, However, you can't use it to 2 tap unarmored people with 45 magnum damage anymore. (realistically using "hold to attack" gives it around the same full auto fire rate) We can come back to this

**Changelog**
:cl:
- tweak: Changed Vector-B to shoot 9x19mm (or 35 auto).

### DIFF
--- a/Resources/Prototypes/_Floof/Catalog/Fills/Crates/armory.yml
+++ b/Resources/Prototypes/_Floof/Catalog/Fills/Crates/armory.yml
@@ -2,7 +2,7 @@
   id: CrateArmoryVector
   parent: [ CrateWeaponSecure, BaseSecurityContraband ]
   name: vector crate
-  description: Contains a Gen-2 Vector with two fourty round mags. Requires Armory access to open.
+  description: Contains a Gen-2 Vector with two thirty round mags. Requires Armory access to open.
   components:
   - type: EntityTableContainerFill
     containers:
@@ -10,5 +10,5 @@
         children:
         - id: WeaponSubMachineGunVector
           amount: 1
-        - id: MagazineMagnumSubMachineGun
+        - id: MagazinePistolSubMachineGun
           amount: 2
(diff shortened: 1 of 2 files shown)


JFerrix (changes requested):
While this is certainly good, I would like to point out that the dual-buff of making it laser-beam accurate (min wielded angle would be fucking 3° and maximum would be 15° at 10 RPS) would absolutely break it. 
Just for comparison, the c20 has a firerate of 8, a mininum wielded angle of 2° and a maximum of 16°. You would literally turn the vector into a c20 on roids.
I do not think that it should be like this. While perhaps realistic, this is something that would completely break balancing.
I suggest reverting the firerate back to 8 and increasing the min angle to 30 and the max angle to 55. Or something like that. Security is already strong enough as is, they do not need and shouldnt have a better c20 simply for the funnies

Raikyr0:
> While this is certainly good, I would like to point out that the dual-buff of making it laser-beam accurate (min wielded angle would be fucking 3° and maximum would be 15° at 10 RPS) would absolutely break it. Just for comparison, the c20 has a firerate of 8, a mininum wielded angle of 2° and a maximum of 16°. You would literally turn the vector into a c20 on roids. I do not think that it should be like this. While perhaps realistic, this is something that would completely break balancing. I suggest reverting the firerate back to 8 and increasing the min angle to 30 and the max angle to 55. Or something like that. Security is already strong enough as is, they do not need and shouldnt have a better c20 simply for the funnies

At your suggestion, I've increased the spread and fixing the fire rate, the Vector's main downside is, it costs MORE then the NT-3 Heavy Machine Gun.

PR #728, opened 2026-06-21
Cooking Rebaked Part 1: Extending Recipes and Adding Food Buffs

SyaoranFox:
## About the PR

This adds the ability for recipes to have multiple outputs, and to indicate how many outputs you want. Recipes can also output fluids. Fluids will be put into a container first if there is enough with room, and then the rest will spill onto the floor.

## Why / Balance
Cooking has always been limited by only having singular outputs. Now recipes can be used to have multiple outputs which also can include biproducts. Completed food should have some kind of benefit beyond vending machine counterparts. The guidebook also now shows the correct machine needed beyond just a microwave.

## Technical details
Results are now a list of IDs, and ResultReagents are a nulled solution that allows for optional liquid outputs for recipes. There is a new FoodEffects component that can be used to add extra effects to completely eating food.

## Media
testRecipy
exampleFoodEffect
cookingRebasedMulti

</p></details>

## Requirements

## Licensing:

## Breaking changes

Recipes now can output one, or many entities. In order to output many of the same entity, it must be put into the results multiple times. They may also accept one or many fluids as part of the recipe results. FoodEffects may also be applied to food to allow for extra buffs after finishing a dish.

The guidebook will also indicate how many entities will be created, but it looks pretty bad. In future someone should make a dictionary to count how many of each entity is made, and then add a number to say how many.

**Changelog**

:cl:
- add: The ability for recipes to indicate many entity IDs, and amounts.
- add: The ability for recipes to produce reagents.
- add: The ability for machines to fill liquid containers with reagents, while also including overflow.
- add: Food may now have innate benefits from eating them. Added with FoodEffect.
- add: Guidebook now supports showing multiple outputs as part of a recipe.
- tweak: How recipes can be built. Results can have multiple outputs.
- fix: RecipeBiscuitSimple now produces 4 simple biscuits instead of 1.
- fix: RecipeHotX now produces 2 hot cross buns instead of 1.
- fix: RecipeMeatBun now produces 2 meat buns instead of 1.
- fix: RecipeSausageRaw now produces 3 raw sausages instead of 1
- fix: Guidebook now indicates the correct machine needed to make a recipe.


Dunrab:
the hot dog water empire will rise

Content.Client/Guidebook/Controls/GuideMicrowaveEmbed.xaml.cs
```diff
@@ -79,14 +81,64 @@ public bool TryParseTag(Dictionary<string, string> args, [NotNullWhen(true)] out
...
+                resultNameMsg.AddMarkupOrThrow(Loc.GetString("guidebook-microwave-solid-name-display",
+                    ("ingredient", entity.Name)));
 
-        RepresentedPrototype = entity;
+                var resultNameLabel = new GuidebookRichPrototypeLink();
+                resultNameLabel.SetMessage(resultNameMsg);
+                ResultsGrid.AddChild(resultNameLabel);
+            }
+
+        }
```
Mnemotechnician, line 149: This code is so awful exactly because you're trying to mix solids and liquids. Making separate lists for both would make the code clearer and yaml type-safer.

Content.Server/Kitchen/EntitySystems/MicrowaveSystem.cs
```diff
@@ -755,7 +761,54 @@ public override void Update(float frameTime)
...
+                                }
+
+                            }
+                            else
+                            {
+                                for (int j = 0; j < result.Value; j++)
+                                {
+                                    Spawn(result.Key, coords);
+                                }
+                            }
```
Mnemotechnician: Same here

Content.Shared/_NF/Medical/SharedMedicalGuideDataSystem.cs
```diff
@@ -75,7 +75,7 @@ public MedicalRecipeData(FoodRecipePrototype proto)
     {
         Identitier = proto.Name;
         Recipe = proto.ID;
-        Result = proto.Result;
+        Result = "proto.Result";
```
Mnemotechnician: ???

Content.Shared/Kitchen/MicrowaveMealRecipePrototype.cs
```diff
@@ -30,8 +30,9 @@ public sealed partial class FoodRecipePrototype : IPrototype
         [DataField("solids", customTypeSerializer: typeof(PrototypeIdDictionarySerializer<FixedPoint2, EntityPrototype>))]
         private Dictionary<string, FixedPoint2> _ingsSolids = new ();
 
-        [DataField("result", customTypeSerializer: typeof(PrototypeIdSerializer<EntityPrototype>))]
-        public string Result { get; private set; } = string.Empty;
+        //Euph, make recipes be able to handle multiple results from a single recipe
+        [DataField("result", customTypeSerializer: typeof(PrototypeIdDictionarySerializer<FixedPoint2, EntityPrototype>))]
+        private Dictionary<string,FixedPoint2> _ingsResult = new ();
```
Mnemotechnician: Changing this to fixedpoint2 would be unnecessary if you had just introduced an additional field ("spawn 0.314 of FoodBread when this meal is cooked" doesn't make any sense)

You could also avoid using a dictionary and use an array instead, and keep this field as an array of entity prototype ids with AbstractPrototypeIdArraySerializer<EntityPrototype> as the serializer. This would allow you to specify results both as
```yaml
results: FoodBread
```
to spawn one bread and as
```yaml
results:
- FoodBread
- FoodBread
- FoodBread
```
to spawn three (or more).

This would also make this PR like 100 times shorter.

Content.Shared/Nutrition/EntitySystems/IngestionSystem.cs
```diff
@@ -379,6 +382,23 @@ private void OnEatingDoAfter(Entity<BodyComponent> entity, ref EatingDoAfterEven
...
+            var time = foodEffect.Time;
+            var mode = foodEffect.Mode;
+            var delay = foodEffect.Delay;
+
+            if (!_effects.HasStatusEffect(entity, effect))
+            {
+                if (mode == StatusEffectMetabolismMode.Update)
+                    _effects.TryUpdateStatusEffectDuration(entity, new EntProtoId(effect), time, delay);
+            }
+        }
```
Mnemotechnician: Listen on FullyEatenEvent on the food item instead of hacking into the injestion system

Mnemotechnician (changes requested):
I think in its current state, the PR would have disasterous consequences for future upstream merges.

Mnemotechnician:
Also, pretty much all the changes to meal prototypes will become redundant if you avoid using a dictionary in MicrowaveMealPrototype. See the comments above.

Content.Shared/_NF/Medical/SharedMedicalGuideDataSystem.cs
SyaoranFox: Oh, yeah, this was a placeholder I forgot to switch over to doing multiple results.

Content.Client/Guidebook/Controls/GuideMicrowaveEmbed.xaml.cs
SyaoranFox: Split into solution and results.

Mnemotechnician (approved):
Looks much better now

PR #740, opened 2026-06-24
Fox Form Updates

Dunrab:
## About the PR
Makes kitsunes fox forms have moving legs and wagging tails when they move around like basemobfox.

## Why / Balance
I thought it would be cool so took the time to get it working. Now with that out of the way some players wanted fox form to be able to pull objects and to have night vision, so i gave the comps however, I don't have a side in the matter so once the discussion is done let me know if we want to remove the comps or keep them.

## Technical details
Adds a movement layer to kitsunes fox form that uses a few new sprites. I was able to make the fox form almost identical to the old one, only large difference is the tail dose not have the white coloring, its the full color of the characters eyes which to me isn't a big deal but may be to others.

## Media
https://github.com/user-attachments/assets/aae805ee-9a5a-46ee-adf2-877473299848

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- add: Kitsune fox forms move their legs and wag their tails when walking.


Dunrab:
media attached, going to let tests run

Mnemotechnician (approved):

PR #758, opened 2026-07-01
No more species req for proto languages

Aureuix:
## About the PR

Removes the species requirement to pick proto languages.
## Why / Balance

Allows for more freedom when creating characters with non standard species appearances (such as humans with fox ears getting vulpini) and allows more opportunity for those with proto languages to use them (some are oh so rare...). Media omitted due to the PR not really changing anything visually. Can attach some at request.

## Requirements

## Licensing:

**Changelog**

:cl:
- tweak: Changed proto languages to not be species locked.

### DIFF
--- a/Resources/Prototypes/_Floof/Traits/Languages/additional_languages.yml
+++ b/Resources/Prototypes/_Floof/Traits/Languages/additional_languages.yml
@@ -88,7 +88,7 @@
   description: trait-FoxLanguage-desc
   category: Languages
   cost: 1
-  #conditions: Euphoria, removing language species reqs to allow for better roleplay and creation of custom species.
+  #conditions:
   #- !type:OneOfSpeciesCondition
   #  species:
   #  - Vulpkanin
@@ -107,7 +107,7 @@
   category: Languages
   cost: 1
   #conditions:
-  #- !type:OneOfSpeciesCondition Euphoria, removing language species reqs to allow for better roleplay and creation of custom species.
+  #- !type:OneOfSpeciesCondition 
   #  species: [Vulpkanin]
   effects:
   - !type:ModifyLanguagesTraitEffect
@@ -123,7 +123,7 @@
   description: trait-CatLanguage-desc
   category: Languages
   cost: 1
-  #conditions: Euphoria, removing language species reqs to allow for better roleplay and creation of custom species.
+  #conditions: 
   #- !type:OneOfSpeciesCondition
     #species:
     #- Felinid
@@ -142,7 +142,7 @@
   description: trait-MonkeyLanguage-desc
   category: Languages
   cost: 1
-  #conditions: Euphoria, removing language species reqs to allow for better roleplay and creation of custom species.
+  #conditions: 
   #- !type:OneOfSpeciesCondition
   #  species:
   #  - Human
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
I'm against this for reason I've expressed many times before.

kivdon:
It's just the proto languages, so it seems fine to me. A lot of the lore and reasoning behind them is questionable or non-existent anyway. Like why are humans able to speak Kobold? How are humans able to speak Simian? 
Barely any of them are ever used anyway. Have you ever seen _anyone_ speak Simian?

Raikyr0:
> Have you ever seen _anyone_ speak Simian?

Ironically, I think I only know 1 character with Simian, and it's Ed Chadderton.

ScarlettJFG:
I personally think *unlocking* languages is a bit of a stretch, why not fix the issues with the current wording for them so that descriptions make sense, or restrict those with wording that makes having them make no sense (body language requiring tails for example), so consider me on the fence with this one...

Raikyr0:
Honestly, given some players give their characters tails or ears, it wouldn't be too much to unlock em.

kivdon:
> I personally think _unlocking_ languages is a bit of a stretch, why not fix the issues with the current wording for them so that descriptions make sense, or restrict those with wording that makes having them make no sense (body language requiring tails for example), so consider me on the fence with this one...

Those are some fair considerations. I would point out, though, that we allow custom species and we have lots of markings to do so with. So if you, for example, were to block humans from taking languages that require a tail, then someone's "human" which is a custom species with a tail (like an ariral I saw) can't take it despite having a tail. Blocking species from things is complicated when we support custom species. 

I would also mention that this isn't really for any languages that "matter". This isn't unlocking the Shadekin language, or the Vox language, or the Asakim language, or anything like that. 
This change will allow more people to potentially communicate with Laika, or with the kobolds that typically get used for meat. This change won't even allow people to speak with Pun Pun though, since I think he still has the AccentReplacement thing.

ScarlettJFG:
> > I personally think _unlocking_ languages is a bit of a stretch, why not fix the issues with the current wording for them so that descriptions make sense, or restrict those with wording that makes having them make no sense (body language requiring tails for example), so consider me on the fence with this one...
> 
> Those are some fair considerations. I would point out, though, that we allow custom species and we have lots of markings to do so with. So if you, for example, were to block humans from taking languages that require a tail, then someone's "human" which is a custom species with a tail (like an ariral I saw) can't take it despite having a tail. Blocking species from things is complicated when we support custom species. 
> 
> I would also mention that this isn't really for any languages that "matter". This isn't unlocking the Shadekin language, or the Vox language, or the Asakim language, or anything like that. 
> This change will allow more people to potentially communicate with Laika, or with the kobolds that typically get used for meat. This change won't even allow people to speak with Pun Pun though, since I think he still has the AccentReplacement thing. 

So then we should make a system so that if you have the markings the species needs, you can take that language, at least for ones like caniluntz. At least that is what I think, I know it'll be a pain to do though...

Aureuix:
if #867 goes through this will be even more relevant. no point to having some of these traits locked if the only species that could even take them get them for free now.

Aureuix:
bumping this since its been a few months

kivdon:
We receive CHelps for giving these languages to people _constantly_, and as far as I can tell, we grant it every single time. 
We might as well allow it as traits so that we can skip that middle step, and so people can make the characters that they really want to make.

PR #768, opened 2026-07-06
Fix Capguns taking other Sidearms as ammo

Dunrab:
## About the PR
As the title states, you can put other sidearms into capguns as ammo. Part of me wants to keep this as a feature, since its harmless and makes the gun into a capgun round but we should probably fix this since you can hide six guns inside of a capgun.

## Technical details
For whatever reason, delta-v gave the RevolverAmmoProvider whitelist the ConcealCarry tag, which all side arms hvae so that they can put into boots (some melee weapons as well like the combat knife.) By doing this they allowed guns to be used as ammo...

## Media

Example Media Embed
If i really need to include media for this i will but im to tired to do that right now.
</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- add: Added fun!
- remove: Removed fun!
- tweak: Changed fun!
- fix: Fixed fun!

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Fun/weapons.yml
+++ b/Resources/Prototypes/Entities/Objects/Fun/weapons.yml
@@ -110,7 +110,7 @@
       tags:
       - CartridgeCap
       - SpeedLoaderCap
-      - ConcealCarry # DeltaV - can be stowed in military boots for now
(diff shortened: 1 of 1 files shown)


SyaoranFox:
Maybe we should change it so you can just fire a gun at someone with a gun? Then that gun can accidently fire another gun filled with guns.

Eh, this works for now.

PR #790, opened 2026-07-12
Hallucination Adjustments

Raikyr0:
## About the PR
Adjusts Frezon to have 50% less duration (down from adding 500 seconds per half unit) and Haloperidol from 10 seconds to 30.

## Why / Balance
Frezon IS making people high for the whole shift, and probably the NEXT shift, I've halved it so -> Haloperidol was buffed so you don't need to drink 3 liters to remove Frezon.

**Changelog**
:cl:
- tweak: Changed Frezon to make you high for half the time, and Haloperidol to be 3 times as effective against being high.

### DIFF
--- a/Resources/Prototypes/Reagents/gases.yml
+++ b/Resources/Prototypes/Reagents/gases.yml
@@ -431,7 +431,7 @@
         minScale: 1
         effectProto: StatusEffectSeeingRainbow
         type: Add
-        time: 500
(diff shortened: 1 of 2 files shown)


Mnemotechnician (approved):

Raikyr0:
Check Failure is just Lottery Box. So, Unrelated.

PR #792, opened 2026-07-12
Allow wearing translators in the ID card slot for people who use collar IDs

DuskyJay:
## About the PR

Very small change, just allows the translator to be worn in the ID slot as an alternative to the neck slot for people who use collar IDs

## Why / Balance

Makes it easier to use a translator while using a collar ID and makes the ID slot not entirely useless for people with ID collars instead.

## Technical details

Tiny yml change, not much else.

## Media
N/A
## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- tweak: Translators can now be used from the PDA slot for people who use their collar as ID!


Dunrab:
good idea

Mnemotechnician (approved):

PR #802, opened 2026-07-14
Funny blood traits

Raikyr0:
## About the PR
This PR adds 3 more blood replacements, Laughter, Nothing, Xenos Blood (long acid name).

## Why / Balance
Flavor.

**Changelog**
:cl:
- add: Adds 3 more blood type traits.


SyaoranFox:
While these may be funny, we cant have useful fluids being in someone's blood like this. Its too easy for someone to abuse this.

Raikyr0:
> While these may be funny, we cant have useful fluids being in someone's blood like this. Its too easy for someone to abuse this.

Which ones would be considered useful?
The one with the most use? would be xenos blood since you could use it to spray acid on people.

PR #807, opened 2026-07-15
Fix Inescapable Goliath Stunlock

kivdon:
## About the PR

Fixed an issue (some called it a glitch) where the Goliath can keep someone stunned forever.
Did so by decreasing the Goliath attack rate by 33%.
Increased Goliath damage by 33% to compensate. 

## Why / Balance

Here's how the Goliath stunlock works: 
You get hit by the Goliath tentacles (very easy to do with latency and the surprisingly large hitbox). You are stunned. 
The Goliath walks over and melees you while you're stunned. 
After 5 seconds of stun, you automatically try to stand. 
The Goliath hits you, preventing you from standing, resetting the doafter. 
The Goliath does this until it can use its tentacles again, which may happen after the first hit or after the second. 
Since you were unable to stand and get away, you're stunned again. 
Process repeats until you're dead. 

tldr: Goliath stuns you, then melees you to prevent you from standing, then stuns you again, then melees you to prevent you from standing, then stuns you again, etc.

Some people have quit Salvage over this, as it's seen as unfair and buggy. 

I decreased Goliath attack speed by 33%. So now, as long as you move while you're trying to stand, you should be able to avoid the first or second melee attack while you're standing up. If you don't move, you will still get stunlocked. And when you do stand, you still have to avoid the tentacles that are about to attack you in 1 second. 
Escaping the Goliath is still very challenging, but at least it's now possible. 

To compensate for decreasing attack speed by 33%, I increased damage by 33%. The Goliath "DPS" theoretically remains unchanged, thus maintaining the difficulty of fighting one. 

## Technical details

Just yml changes.

## Media

image
This screenshot shows the changes made to the melee.

I'd include videos of the stunlock before and after changes, but I don't have a way of recording. I can easily show in-game by editing the component, if anyone wants to see it. 

</p></details>

## Requirements

## Licensing:

  - [ ] AGPL (https://github.com/Floof-Station/Panta-Rhei/blob/master/LICENSE-AGPLv3.txt) 
    

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Goliath stunlock by decreasing melee attack rate by 33%. Increased damage by 33% to compensate.


kivdon:
Tested.

Mnemotechnician (approved):

SyaoranFox (approved):
I suspect this means that characters with lower HP will have a harder time against Goliaths. We shall see.

PR #422, opened 2026-04-15
Passenger Emotional Support Collars

Dunrab:
## About the PR
You ever want to run around as a passenger wtih out using a pda? Well now you can. These collars will let you store you id, pen and act as a pda.

## Why / Balance
Why you ask? because who needs a pda when you can have style and a pda.

## Technical details
This hurt my head and probably menmos who helped me debug why the job icon was not showing up and id renaming.

## Media
<details> <summary><h3>Click to show</h3></summary> <p>

Small show casing of the id, more work needs to be done.

https://github.com/user-attachments/assets/4a3cff43-0215-43d7-bc39-40bd7a23f9c0

Loadout options:

image

</p></details>

## Requirements

- [x] I have tested all added content and changes.
- [x] I have added media to this PR or it does not require an ingame showcase.

## Licensing:

- [x] I confirm that I am the creator of the content in this PR, and allow licensing it under the following license(s), or that the original author has given me permission to do so:
  - [X] AGPL (https://github.com/Floof-Station/Panta-Rhei/blob/master/LICENSE-AGPLv3.txt) 
  - [ ] MIT (https://github.com/Floof-Station/Panta-Rhei/blob/master/LICENSE-MIT.txt)
    

## Breaking changes

**Changelog**

:cl:
- add: Passengers can now select a new type of pda in their loadouts in the form of a collar. These new collars function the same as a pda and can be worn around the neck.
- add: Added the emotional support collars to the hops ptech.

PR #817, opened 2026-07-18
Improve handling of non-ID entities in the ID slot and fix error caused by #792

DuskyJay:
## About the PR

Fixes error caused by oversight in testing of #792 and allows the `IdExaminableSystem` to attempt checking the neck slot for ID if the ID slot contains an invalid ID.

## Why / Balance

While testing #792 I failed to catch an error when wearing a translator in the ID slot that would prevent your collar ID card from being examined due to the ID card slot being read first, even if the contained entity wasn't a valid ID.

These changes allow the `IdExaminableSystem` to check the ID slot, determine if it contains a valid ID and if not then attempt to check the neck slot, allowing for arbitrary items in the ID slot to not prevent examining a collar ID.

## Technical details

Changes to shared `IdExaminableSystem`.
Moves some logic from the `GetInfo` function into a helper `GetIdInfo` function.

Previous `GetInfo` function:
```C#
    public string? GetInfo(EntityUid uid)
    {
        if (_inventorySystem.TryGetSlotEntity(uid, "id", out var idUid) || _inventorySystem.TryGetSlotEntity(uid, "neck", out idUid)) // Floofstation Pet IDs to check neck slot
        {
            // PDA
            if (TryComp(idUid, out PdaComponent? pda) &&
                TryComp<IdCardComponent>(pda.ContainedId, out var id))
            {
                return GetNameAndJob(id);
            }
            // ID Card
            if (TryComp(idUid, out id))
            {
                return GetNameAndJob(id);
            }
        }
        return null;
    }
```

Updated `GetInfo` function:
```C#
    public string? GetInfo(EntityUid uid)
    {
        // Check ID slot
        if (_inventorySystem.TryGetSlotEntity(uid, "id", out var idUid))
        {
            var idInfo = GetIdInfo((EntityUid)idUid);
            if (idInfo != null) return idInfo;
        }
        // Check neck slot
        if (_inventorySystem.TryGetSlotEntity(uid, "neck", out idUid))
        {
            var idInfo = GetIdInfo((EntityUid)idUid);
            if (idInfo != null) return idInfo;
        }
        return null;
    }
```

New `GetIdInfo` helper function:
```C#
    private string? GetIdInfo(EntityUid uid)
    {
        // PDA
        if (TryComp(uid, out PdaComponent? pda) &&
            TryComp<IdCardComponent>(pda.ContainedId, out var id))
        {
            return GetNameAndJob(id);
        }
        // ID Card
        if (TryComp(uid, out id))
        {
            return GetNameAndJob(id);
        }
        return null;
    }
```

## Media

### Before:
Translator blocks collar ID:
image

### After:
Translator does not block collar ID:
image

Clown ID takes priority as it is in the ID slot:
image

Clown ID functions normally when neck slot contains non-ID:
image

Functions fine with nothing in ID slot:
image

Functions fine with nothing in neck slot:
image

Collar ID remains visible with Throngler in ID slot:
image
</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- fix: Fixed error that made it impossible to examine the collar ID of someone wearing a translator in their ID slot.

### DIFF
--- a/Content.Shared/Access/Systems/IdExaminableSystem.cs
+++ b/Content.Shared/Access/Systems/IdExaminableSystem.cs
@@ -46,25 +46,40 @@ public sealed class IdExaminableSystem : EntitySystem
         return GetInfo(uid) ?? Loc.GetString("id-examinable-component-verb-no-id");
     }
 
+    // Start Floofstation
     public string? GetInfo(EntityUid uid)
     {
-        if (_inventorySystem.TryGetSlotEntity(uid, "id", out var idUid) || _inventorySystem.TryGetSlotEntity(uid, "neck", out idUid)) // Floofstation Pet IDs to check neck slot
+        // Check ID slot
+        if (_inventorySystem.TryGetSlotEntity(uid, "id", out var idUid))
         {
-            // PDA
-            if (TryComp(idUid, out PdaComponent? pda) &&
-                TryComp<IdCardComponent>(pda.ContainedId, out var id))
-            {
-                return GetNameAndJob(id);
-            }
-            // ID Card
-            if (TryComp(idUid, out id))
-            {
-                return GetNameAndJob(id);
-            }
+            var idInfo = GetIdInfo((EntityUid)idUid);
+            if (idInfo != null) return idInfo;
+        }
+        // Check neck slot
+        if (_inventorySystem.TryGetSlotEntity(uid, "neck", out idUid))
+        {
+            var idInfo = GetIdInfo((EntityUid)idUid);
+            if (idInfo != null) return idInfo;
         }
         return null;
     }
 
+    private string? GetIdInfo(EntityUid uid)
+    {
+        // PDA
+        if (TryComp(uid, out PdaComponent? pda) &&
+            TryComp<IdCardComponent>(pda.ContainedId, out var id))
+        {
+            return GetNameAndJob(id);
+        }
+        // ID Card
+        if (TryComp(uid, out id))
+        {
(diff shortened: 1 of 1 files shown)


DuskyJay:
First PR that makes actual edits to C# code, wish me luck.

Mnemotechnician (changes requested):
there are like 3 more systems that contain similar hardcoded slot checks. You need to edit them as well, and preferably by utilizing this existing method. See #422

DuskyJay:
I don't really see any issues with the other systems, access works fine and just combines the access of both the collar ID and the ID slot from my testing and the spawning system doesn't seem to require changes either since nothing should be placed in the ID slot  by loadouts unless it is a PDA.

This change primarily aims to fix the issue of inspecting the collar ID of someone who has their ID slot obstructed by another entity.

DuskyJay:
Identity, access logs and paper signing also don't seem to require any changes as they work fine when the ID slot is obstructed anyway.

DuskyJay:
If you could please direct me to what systems require changes as well that would be appreciated but as far as I can see all the systems referenced in #422 shouldn't have any issues

Dunrab:
> If you could please direct me to what systems require changes as well that would be appreciated but as far as I can see all the systems referenced in #422 shouldn't have any issues

i had to make a fix to #422 look at #433 as well incase

DuskyJay:
It doesn't look like the changes required in #433 should cause any issues with this

SyaoranFox (approved):

PR #825, opened 2026-07-20
Hair Brush

Dunrab:
## About the PR
This pr does two things:

1. Adds a new pop up component/system for entities so that they can have custom pop up messages when used on self or a target.
2. Adds a hairbrush that uses comp/system.

## Why / Balance
A hair brush was requested so I added one, but while adding it I realized that there was not a good way to have it have it use a different pop up message so at first I was going to make a component that would just be used for the brush its self. That quickly changed after Speaking with Menmo who suggested turning into a system instead that is generic so it can be used for other items as well so here we are.

See the brush.yml to see how to use this comp, but its pretty standard. If anyone plans to add more brush variants, they can parent off of the brush i made.

## Technical details
Adds a new component called PopUpOnUseComponent that allows for custom pop up messages for self and targets along with whitelists, blacklists and the ability to have a sound play on doafter.

## Media

https://github.com/user-attachments/assets/4189c32b-6a18-44b3-878e-cb3c815683e9
</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- add: Hair brushes have been added to Loadouts and the Pet Shop.


Content.Shared/_Floof/Popup/PopUpOnUseSystem.cs
```diff
@@ -0,0 +1,78 @@
...
+        _whitelist.IsWhitelistPass(ent.Comp.Whitelist, target) &&
+        !_whitelist.IsWhitelistPass(ent.Comp.Blacklist, target);
+
+    public override void Initialize()
+        {
+            base.Initialize();
+
+            SubscribeLocalEvent<PopUpOnUseComponent, AfterInteractEvent>(OnAfterInteract);
+            SubscribeLocalEvent<PopUpOnUseComponent, PopUpOnUseDoAfterEvent>(OnDoAfter);
+        }
```
Mnemotechnician: Formatting, also calling the superinitializer is redundant
```suggestion
      {
          SubscribeLocalEvent<PopUpOnUseComponent, AfterInteractEvent>(OnAfterInteract);
          SubscribeLocalEvent<PopUpOnUseComponent, PopUpOnUseDoAfterEvent>(OnDoAfter);
      }
```

Content.Shared/_Floof/Popup/PopUpOnUseSystem.cs
```diff
@@ -0,0 +1,78 @@
...
+        BeginDoAfterEvent(ent, args.Target.Value);
+        args.Handled = true;
+
+    }
+
+    public void BeginDoAfterEvent(Entity<PopUpOnUseComponent> ent, EntityUid target)
+    {
+        if (!Resolve(ent, ref ent.Comp!, false))
+            return;
+    }
```
Mnemotechnician: this method does nothing, might just remove it

```suggestion
        args.Handled = true;
    }
```

Mnemotechnician (approved):
other than that it's fine

Content.Shared/_Floof/Popup/PopUpOnUseSystem.cs
Dunrab: yeah its a remnant of when i was doing just the brush comp, clearly forgor to remove it

SyaoranFox (approved):

PR #830, opened 2026-07-21
Ports the Neuralyzer & its trait from HL

DoubleHypercube:
## About the PR

The Neuralyzer is designed for hypno/mindcontrol/brainwashing/etc kink play; as well as a Brainwasher trait that does the same thing but doesn't require an item.

## Why / Balance

A _lot_ of people want mechanical backing for hypnosis or related kink play, and this provides Exactly That.
The original HL implementation, I will say, made mindshields block neuralyzers and had the trait cost points. I've opted to not port that as I don't think that's the direction we should be or even want to take; HL did it only because neuralyzers (iirc) were valid for antag use and I don't believe we have the same philosophy nor do I think excluding Command and Security from hypnosis kink play is fair.

## Technical details

New systems to handle it, of course.

## Media

image
image</details>

## Requirements

## Licensing:

image

  - [] MIT (https://github.com/Floof-Station/Panta-Rhei/blob/master/LICENSE-MIT.txt)
    

## Breaking changes

**Changelog**

:cl: DoubleHypercube, Notarin
- add: Ported the brainwashing systems from HL to Euphoria!


Cprn-Sprkl:
I'd recommend changing the language in the pop up to better fit Euphoria.

image

This section is a sticking point since their rules are not our rules. We certainly don't want to imply that borg laws are server rules, and wouldn't want that for compulsions either.

image

This is heavily restrictive to how people can RP their own hypnosis. Since this won't be gameplay focused here there doesn't need to be any hard definition of how it works for the person being compelled. Offering suggestions might help for those new to it, but anything saying "it is this" is more restrictive than we'd typically do here.

If you want suggestions on what language would be good ask on the Hypnosis thread in the discord and you'll probably get lots to choose from.

The bottom P.S. is excellent and should stick around in any iteration of this pop up.

ASaltySeabass:
> I'd recommend changing the language in the pop up to better fit Euphoria.
> 
> image
> This section is a sticking point since their rules are not our rules. We certainly don't want to imply that borg laws are server rules, and wouldn't want that for compulsions either.
> 
> image
> This is heavily restrictive to how people can RP their own hypnosis. Since this won't be gameplay focused here there doesn't need to be any hard definition of how it works for the person being compelled. Offering suggestions might help for those new to it, but anything saying "it is this" is more restrictive than we'd typically do here.
> 
> If you want suggestions on what language would be good ask on the Hypnosis thread in the discord and you'll probably get lots to choose from.
> 
> The bottom P.S. is excellent and should stick around in any iteration of this pop up.

Seems like a holdover from HL, where the neuralyzer/hypno/mind control system is used for antag related purposes, something that doesn't really apply here.

DoubleHypercube:
Yeah, I may end up just removing the popup in its entirety now that it's mentioned - it was written for a different server and most of it doesn't apply here. I can't really think of a way to reword it that doesn't result in making it small enough to be removed outright.

DoubleHypercube:
~~I appear to have fucked it. Please hold~~

I have successfully unfucked it. Please ignore the horrible commit history and supposed RSI changes.

PR #831, opened 2026-07-21
Skip preloading lobby art to save >200MB VRAM (and also shrink Ratvar in the wash)

DoubleHypercube:
## About the PR

It does what it do. Also makes CosCult map not always load.

## Why / Balance

Saves over *300MB* of VRAM.
Also might help server performance from not loading a map it doesn't need? Dunno.

## Media
Live vs Savings:
image

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl: DoubleHypercube
- tweak: Skipped preloading unneeded lobby arts to save a significant portion of VRAM.
- tweak: Shrunk Ratvar in the wash to halve his size so he doesn't take up 36MB VRAM.
- remove: The Cosmic Cult pocket dimension is no longer loaded on non-CosCult rounds.


Cprn-Sprkl:
~~This is visibly reducing the quality of the images. Request a properly re-scaled image from the artists if the savings is significant enough for this to be needed. One of these images is being fully squashed into a different ratio which significantly alters the art.~~

This is no longer applicable with the change in approach.

SyaoranFox (approved):

Dunrab (commented):
You are touching upstream parallax that our maps use show screenshots of the backgrounds not looking terrible.

DoubleHypercube:
> You are touching upstream parallax that our maps use show screenshots of the backgrounds not looking terrible.

I'm... not? I'm just making them `preload: false`. This method doesn't physically change the texture, it just tells the engine to not load it into VRAM unless it *needs* to display it.

Dunrab:
> > You are touching upstream parallax that our maps use show screenshots of the backgrounds not looking terrible.
> 
> I'm... not? I'm just making them `preload: false`. This method doesn't physically change the texture, it just tells the engine to not load it into VRAM unless it _needs_ to display it.

Coscult pocket dimension is loaded every round. I'm not going to argue with you.

DoubleHypercube:
> Coscult pocket dimension is loaded every round.

~~Is there a reason for this? Surely we could skip loading it if it's not a coscult round...?~~
Yes to both. It's technically loaded because of "Rogue Ascendants", whatever those are - but as far as I can tell (`grep -ir rogue`, `grep -ir ascend`) we don't have those. So, I've moved it to a LoadMapRule same as how Nukeops doesn't preload the nukie base. _Also_, apparently the main menu has a parallax too for some reason? So by not loading that, we can save a few more MB - this is literally (as far as I can tell) only used in the _menu_ menu, which is only ever seen in a development environment (because it doesn't autoconnect to a server) or when you disconnect without outright quitting which as far as I'm aware literally noone does.

> I'm not going to argue with you.

There's no need for hostility.
(Also I'm not even arguing...)

Dunrab:
> > Coscult pocket dimension is loaded every round.
> 
> Is there a reason for this? Surely we could skip loading it if it's not a coscult round...?
> 
> > I'm not going to argue with you.
> 
> There's no need for hostility. (Also I'm not even arguing...)

Cosults maker or our upstream (not sure which) made it load every round to prevent metagaming to my knowledge, im assuming they had a problem with it since if it doesn't load every round someone can ghost to verify if its a coscult round or not

DoubleHypercube:
Yeah it turns out that apparently it was only loaded for one midround antag which I don't even think we _have_ lol
I did check if metagaming was the case but I can't warp to it anyway even though it's definitely loaded - this might be because it's loaded as paused? 

`// Rogue Ascendants use this too, which are generalized MidRoundAntags, so we keep the map around. If you're porting cosmic cult, and do not want rogue ascendants, feel free to move this into selective usage akin to NukeOps base.` was the comment above `CosmicCultSystem#OnRoundStart`

image
image

Dunrab:
> Yeah it turns out that apparently it was only loaded for one midround antag which I don't even think we _have_ lol I did check if metagaming was the case but I can't warp to it anyway even though it's definitely loaded - this might be because it's loaded as paused?
> 
> image image

id imagine its paused till its used? It may also lack a regular ghost warp point, not sure i haven't ever looked at the map but i don't think we have any additional antags running for it other than the game mode its self so should be fine

Content.Server/_DV/CosmicCult/CosmicCultSystem.cs
```diff
@@ -120,11 +120,11 @@ public void MalignEcho(Entity<CosmicCultComponent> uid)
...
-    private void OnRoundStart(RoundStartingEvent ev)
-    {
-        if (_mapLoader.TryLoadMap(_mapPath, out var map, out _, new DeserializationOptions { InitializeMaps = true }))
-            _map.SetPaused(map.Value.Comp.MapId, false);
-    }
+    // private void OnRoundStart(RoundStartingEvent ev) // Euphoria: AFAICT, we don't have rogue ascendants - couldn't find them in the codebase, so we don't need to load this.
+    // {
+    //     if (_mapLoader.TryLoadMap(_mapPath, out var map, out _, new DeserializationOptions { InitializeMaps = true }))
+    //         _map.SetPaused(map.Value.Comp.MapId, false);
+    // }
```
Mnemotechnician, line 128: Isnt this where minds are sent to when a cultist uses the stun action? You need to set it after the map gets loaded, which you dont seem to be doing.

DoubleHypercube: I mean, if there are cultists, wouldn't it be a cult round (which would trigger LoadMapRule)? And IIRC this is only used for the finale, at least that's what the parallax ID implies.

Mnemotechnician: actually nevermind, i thought the system was saving the mapid of that map. it isn't.

Mnemotechnician (approved):

PR #835, opened 2026-07-21
Don't Load Ghostbar in Debug

Dunrab:
## About the PR
Don't load the ghostbar in debug, can manually load it if needed.

## Why / Balance
Much like lavaland, no one really needs this to load except when working with it so why make people load it in the first place?

## Technical details
Small changes to GhostBarSystem

## Requirements

## Licensing:

### DIFF
--- a/Content.Server/_Goobstation/Ghostbar/GhostBarSystem.cs
+++ b/Content.Server/_Goobstation/Ghostbar/GhostBarSystem.cs
@@ -57,8 +57,8 @@ public sealed class GhostBarSystem : EntitySystem
         return;
         if (_mapLoader.TryLoadMap(_mapPath, out var map, out _, new DeserializationOptions { InitializeMaps = true }))
             _mapSystem.SetPaused(map.Value.Comp.MapId, false);
-    }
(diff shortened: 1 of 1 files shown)


Content.Server/_Goobstation/Ghostbar/GhostBarSystem.cs
```diff
@@ -57,8 +57,8 @@ private void OnRoundStart(RoundStartingEvent ev)
         return;
         if (_mapLoader.TryLoadMap(_mapPath, out var map, out _, new DeserializationOptions { InitializeMaps = true }))
             _mapSystem.SetPaused(map.Value.Comp.MapId, false);
-    }
         #endif // we don't want to load the ghostbar in debug
+    }
```
DoubleHypercube, line 61: I've never really used preprocessor directives but shouldn't the `#endif` be directly after the ret?

Mnemotechnician: yes

Mnemotechnician (approved):

SyaoranFox (approved):

PR #838, opened 2026-07-21
Adds Ion repair to cables

Raikyr0:
## About the PR
This PR adds Ion Damage repair to cables.

## Why / Balance
This PR aims to fix the issue which plagues cybernetics being un-repairable and unusable as organic species don't have natural ion regeneration. Also makes it so IPCs can get up from Ion damage faster should someone recover them.

**Changelog**
:cl:
- add: Added Ion damage repair to cables, you may now fix cybernetic limbs which don't naturally regenerate instead of replacing them, IPCs may also be repaired so Ion damage goes away faster.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Tools/cable_coils.yml
+++ b/Resources/Prototypes/Entities/Objects/Tools/cable_coils.yml
@@ -43,6 +43,7 @@
         Heat: -3.22 # DeltaV was -1.66
         Shock: -3.22 # DeltaV was -1.66
         Cold: -3.22 # DeltaV
+        Ion: -5 # Euphoria - This is mostly a fix for organics with cybernetics which don't autoheal but works as a way to get someone EMP'd up faster
   # EE Change End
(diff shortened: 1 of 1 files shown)


Mnemotechnician (changes requested):
Silicons can receive ion damage too, 5/use is wayyy too much healing for this

Raikyr0:
> Silicons can receive ion damage too, 5/use is wayyy too much healing for this

I adjusted it to heal as much as it would heal burn.
And yes I know Silicons can receive Ion damage, however they have no way to expedite healing it other then waiting for ion to naturally tickdown (made worse by the fact they will MOST likely be dead the whole time), so this also resolves that.

SyaoranFox:
Going to put a hold on this. Please test this PR after the reimplementation of IPCs come in.

SyaoranFox:
Has this been tested against new IPCs?

Mnemotechnician:
it wasn't, ion damage is gone

PR #840, opened 2026-07-22
Language prefixes~!

DoubleHypercube:
## About the PR

Adds typeable prefixes for all languages, like `^taj` for Siik'tajr or `^trd` for Tradeband.

## Why / Balance

FAR easier to type "`^vulYip woof bark`" than it is to press L (or worse, click the menu), select the language you want, type your message, THEN go back to Tau-Ceti or your previous language.

## Technical details

Based on Starlight PR https://github.com/ss14Starlight/space-station-14/pull/3006

## Media
image

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl: DoubleHypercube
- add: Ported language prefixes from Starlight for QoL!


Content.Shared/_Floof/Language/Systems/SharedLanguageSystem.cs
```diff
@@ -168,4 +175,49 @@ internal int PseudoRandomNumber(int seed, int min, int max)
...
+    /// <param name="modifyText">Whether to allow this function to modify the resulting text string or not.</param>
+    /// <returns></returns>
+    public LanguagePrototype GetLanguageFromPrefix(Entity<LanguageSpeakerComponent?> ent, ref string input, out bool parsed, bool modifyText = false)
+    {
+        parsed = false;
+        // Fallback if unable to get the current selected language. Selected language is used if unable to parse.
+        if (!Resolve(ent, ref ent.Comp, logMissing: false)
+            || string.IsNullOrEmpty(ent.Comp.CurrentLanguage)
+            || !_prototype.TryIndex<LanguagePrototype>(ent.Comp.CurrentLanguage, out var proto))
+            return Universal;
```
Mnemotechnician: ai-slop, just call the api method here

```suggestion
        if (!Resolve(ent, ref ent.Comp, logMissing: false))
            return Universal;
        var proto = GetLanguage(ent);
```

Content.Shared/_Floof/Language/Systems/SharedLanguageSystem.cs
```diff
@@ -168,4 +175,49 @@ internal int PseudoRandomNumber(int seed, int min, int max)
...
+        foreach (var langId in ent.Comp.SpokenLanguages)
+        {
+            if (!_prototype.TryIndex(langId, out var lang))
+                continue;
+            if (lang.ChatPrefix is null)
+                continue;
+            if (lang.ChatPrefix.Length != 3)
+            {
+                throw new Exception(
+                    $"Chat prefixes must be 3 characters long. {lang.Name}'s prefix is {lang.ChatPrefix}");
```
Mnemotechnician: Another case of ai-slop. Throwing an exception here will result in undefined behavior of the chat controller. Log an error and return Universal instead. Also, why the hell is it hardcoded to be 3 characters long? You could call `text[1..].StartsWith(lang.ChatPrefix, StringComparison.CurrentCultureIgnoreCase)` and then it'd support arbitrary prefix lengths.

Content.Client/UserInterface/Systems/Chat/ChatUIController.cs
```diff
@@ -68,8 +71,10 @@ public sealed partial class ChatUIController : UIController
     [UISystemDependency] private readonly TransformSystem? _transform = default;
     [UISystemDependency] private readonly MindSystem? _mindSystem = default!;
     [UISystemDependency] private readonly RoleCodewordSystem? _roleCodewordSystem = default!;
+    [UISystemDependency] private readonly LanguageSystem? _lang = default!; // Starlight
 
     private static readonly ProtoId<ColorPalettePrototype> ChatNamePalette = "ChatNames";
+    private static readonly string SendingLanguageLocId = "language-chat-confirmation"; // Starlight: Language prefixes
```
Mnemotechnician: There's no reason to store this locid in this way. Just inline it. The yaml linter won't validate the existence of a locale string with this key because the type of this field is NOT LocId

Resources/Prototypes/_Floof/Language/FactionSpecific/blood_cult.yml
```diff
@@ -1,6 +1,7 @@
 ﻿# Used by followers of Nar'Sie
 - type: language
   id: Eldritch
+  chatPrefix: eld # Euphoria/Starlight
```
Mnemotechnician: You're putting the chatPrefix field in the header of the language prototype, which violates every convention. All prototype-specific fields should go after generic ones (type, id, parent, abstract, ...). 

This needs to be fixed for every language prototype.

Content.Server/Chat/Systems/ChatSystem.cs
```diff
@@ -440,6 +443,9 @@ private void SendEntitySpeak(
 
         var speech = GetSpeechVerb(source, message);
 
+        if (message.StartsWith(SharedLanguageSystem.ChatPrefixChar)) // Starlight: Language prefixes
+            language = _language.GetLanguageFromPrefix(source, ref message, out _, true);
```
Mnemotechnician: You're only modifying the language in SendEntitySpeak. This won't affect whispers, radio messages, and much much more. **Also this doesn't even check if the entity can speak that language, allowing anyone to speak any language!** You should instead edit TrySendInGameICMessage, parse the prefix there, and then call SubstituteEntityLanguage to replace its chosen language for the duration of the method (see https://github.com/Floof-Station/Panta-Rhei/blob/master/Content.Server/_Floof/Language/LanguageSystem.Utility.cs).
```cs
var languageBasedOnPrefix = _language.GetLanguageFromPrefix(source, ref message, out var languagePrefixParsed, true);
if (languagePrefixParsed && _language.CanSpeak(source, languageBasedOnPrefix))
    using var _ = _language.SubstituteEntityLanguage(source, languageBasedOnPrefix.id);
```
(notice that you MUST NOT add braces to this if-statement, otherwise the language substitution will be immediately cancelled)

Mnemotechnician (changes requested):
The only logical conclusion i can draw from the glaring issues with this PR is that starlight ai-generated it. The said issues need to be fixed before we can accept it.

Content.Shared/_Floof/Language/Systems/SharedLanguageSystem.cs
DoubleHypercube: I think the reason a specific prefix length was chosen is because otherwise you'd have to add another character to determine where the prefix ends or try every possible prefix character by character until you find one which might have possible performance issues? not entirely sure though

Mnemotechnician:
Also, please, link the original PR.

Content.Server/Chat/Systems/ChatSystem.cs
DoubleHypercube: I thought `using var` statements had to be in a braced if-statement? Rider throws `Invalid embedded statement` for that at least though I'm probably not doing it right so I'll see what I can do (inserting it in TrySendInGameICMessage, after SanitizeInGameICMessage call)

DoubleHypercube:
PR linked.

Content.Server/Chat/Systems/ChatSystem.cs
DoubleHypercube: Actually I can just manually dispose it (unless that has performance/other implications).

Mnemotechnician: Oh yea, I misremembered. C# doesn't allow auto-disposed variables inside braceless conditional statements. But you should be able the following syntax instead: `using var _ = condition ? SubstituteEntityLanguage(...) : null;`, C# will just do nothing if the auto-disposed variable is null.

So like this:
```cs
var languageBasedOnPrefix = _language.GetLanguageFromPrefix(source, ref message, out var languagePrefixParsed, true);
// Override entity language with the one specified in the prefix if there is one
using var _ = (languagePrefixParsed && _language.CanSpeak(source, languageBasedOnPrefix))
    ? _language.SubstituteEntityLanguage(source, languageBasedOnPrefix.id) : null;
```

DoubleHypercube: That works. Probably better than the manual-dispose workaround I came up with, will implement :+1:

Content.Shared/_Floof/Language/Systems/SharedLanguageSystem.cs
Mnemotechnician: Performance is literally negligible here. Though it *would* cause issues if two languages shared the same common prefix (it's a common problem in information science, tldr the solution is to ensure no prefix string is the prefix of another prefix string). But it's fine as it is.

Mnemotechnician (approved):
Looks mostly fine now... the rest is just nitpicks.

Content.Server/Chat/Systems/ChatSystem.cs
SyaoranFox: Please increment this counter by how many hours you have spent on this.

DoubleHypercube: best effort guesstimate says maybe a couple hours, counter incremented

SyaoranFox (approved):
Thank you for continuing to document the torment of the chat system ^.^

PR #846, opened 2026-07-23
Curve Research Scaling and increase Weight of Glimmer Research Theft

SyaoranFox:
## About the PR

Science points from artifacts now scale from 1 to 1.5 using the following curve/formula.
EpiScaling

Increased the weight for the Glimmer Research Stealer event. Made its message be broadcast through common rather than science

## Why / Balance
Epi has been using super high glimmer as a reason to speed through research as fast as possible without regard to the dangers of high glimmer. Now, if the scaling at high glimmer is much worse, and, in theory, the events should be more devastating.

## Technical details
I implemented the above equation for scaling. I also changed the weight of GlimmerResearchStealer to be 17 rather than the usual 12.

## Media

Example Media Embed

</p></details>

## Requirements

## Licensing:

**Changelog**

:cl:
- tweak: Glimmer scaling for points is now capped at 1.5, down from 2.
- tweak: Glimmer scaling gives reduced returns when approaching 1000 glimmer.
- tweak: Glimmer Research Stealer event should appear more often, and broadcast on common rather than science.
- fix: Glimmer Research Stealer event should trigger now (hopefully).

### DIFF
--- a/Resources/Prototypes/_DV/GameRules/glimmer_events.yml
+++ b/Resources/Prototypes/_DV/GameRules/glimmer_events.yml
@@ -240,6 +240,8 @@
   parent: BaseGlimmerEvent
   id: GlimmerResearchStealer
   components:
+  - type: StationEvent
+    weight: 17
   - type: GlimmerEvent
     minimumGlimmer: 600
     glimmerBurnLower: 25
--- a/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
+++ b/Content.Server/Xenoarchaeology/Equipment/ArtifactAnalyzerSystem.cs
@@ -66,7 +66,7 @@ public sealed class ArtifactAnalyzerSystem : SharedArtifactAnalyzerSystem
         if (sumResearch <= 0)
             return;
 
-        _glimmerSystem.Glimmer += sumGlimmer; // DeltaV - Add glimmer based on extracted points.    
+        _glimmerSystem.Glimmer += sumGlimmer; // DeltaV - Add glimmer based on extracted points.
         _research.ModifyServerPoints(server.Value, sumResearch, serverComponent);
         _audio.PlayPvs(ent.Comp.ExtractSound, artifact.Value);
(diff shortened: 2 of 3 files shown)


Resources/Prototypes/_DV/GameRules/glimmer_events.yml
```diff
@@ -240,6 +240,8 @@
   parent: BaseGlimmerEvent
   id: GlimmerResearchStealer
   components:
+  - type: StationEvent
+    weight: 17
```
Mnemotechnician, line 244: This field is ignored, you need to change the weight of this entity in the glimmer event table, i think.

Aureuix:
as someone who quit playing epi due to boredom, this pr just feels like its going to make the whole epi issue worse. curbing glimmer skill expression will lead to epi being a less interesting job with less to learn. As a player i would love to see more interesting glimmer events (especially at 600+) that arent just "nobody can have fun for the rest of the round" like mass mindswap was rather than just gutting all of the positive effects of glimmer. Just remove glimmer or gut the positive effects entirely if you dont want epi players to play around the mechanic that's specifically intended to be used to boost productivity at a cost. Trying to "make epi stop using high glimmer" is like trying to "make engineering stop running 2 engines at once for ptl" its an essential part of the dept's gameplay loop's skill expression.

Resources/Prototypes/_DV/GameRules/glimmer_events.yml
SyaoranFox: Looking into it, the Event Handlers do consider the weight value. If I were to change the glimmer event table, I would need to add more glimmer events to make other events more predominant. As it is, all events are present once, and only some appear more often than others.

Mnemotechnician: Yeah, actually you made this event part of the global pool by adding a stationevent comp. I'm not sure if that's a good idea.

Mnemotechnician (approved):

PR #850, opened 2026-07-23
Lower diamond sale price

Mnemotechnician:
## About the PR
Sale price of a material is measured in spesos per unit. One sheet of a material is equal to 100 units (due to its MaterialComposition). Diamonds go for 60k/stack because of that (20 spesos/unit * 100 units/item * 30 item = 60 kspesos). Now they'll go for 9k/stack which is far more reasonable given how common they are.

## Why/Balance
Salvage can bring stacks of diamonds from mining expeditions after they get an advanced ore processor. It's absolutely absurd that salvage can generate more money in a single trip than average cargo ever will in an entire round.

**Changelog**
:cl:
- tweak: Lowered the sale price of a stack of diamonds to 9k spesos.


Ryunexo:
On the side of big diamond here, slightly.

My main gripe with this is that I feel like the factor of profits getting split between departments is not being taken into account. While 60K is a lot, even with the sales split - if you lower it to 9K. Selling a full stack of diamonds is going to bring in less funding per department than some bounties. I mean at that point, the lemon bounty sells for 10k, I think diamonds should be a bit more valuable than the produce at a farmer's market.

I think if you just split it so that the collective cost of a stack is 30K. It's enough that it's worthwhile to cash it in but not enough to entirely supply a department's funds, thus making the rest of the bounty loop redundant.
It also better reflects their actual rarity, diamonds aren't rare sure but they're certainly not common. You only tend to see a full stack of diamonds _after_ salvage has already done one or two runs and only _after_ they have an industrial processor as that is the only way to process diamonds afaik.

Mnemotechnician:
> My main gripe with this is that I feel like the factor of profits getting split between departments is not being taken into account.

There's no need to take that into account: the addition of departmental funds and order computers didn't magically make people order more than they did pre-rebase. It's just that now SOME funds go towards departments and those departments sometimes use their "own" funds to order their stuff, which in turn lowers the amount of funds cargo has to take from their account (which btw kills a lot of potential rp, including conflict rp, but that's besides the point), 

Also, 9k/stack is about what we had pre-rebase (it was lowered there too). It used to be 10k/stack (the price of diamonds was set to 3.3333 spesos/unit)

SyaoranFox:
going from 20 to 3 may be a bit harsh, but I have seen at least one logistics department that got 500k by just doing diamonds. Can we try it at 10 first?

Mnemotechnician:
> going from 20 to 3 may be a bit harsh, but I have seen at least one logistics department that got 500k by just doing diamonds. Can we try it at 10 first?

The other day logi made over 1 million by selling a crate of diamonds. Even 3/u might be too much.

hollowarc:
> > going from 20 to 3 may be a bit harsh, but I have seen at least one logistics department that got 500k by just doing diamonds. Can we try it at 10 first?
> 
> The other day logi made over 1 million by selling a crate of diamonds. Even 3/u might be too much.

That kind of haul takes quite some time to actually mine out (often a good portion of the entire shift), and its done because making lots of money is fun. 5-10 is a good compromise.

Mnemotechnician:
> That kind of haul takes quite some time to actually mine out (often a good portion of the entire shift), and its done because making lots of money is fun. 5-10 is a good compromise.

It doesn't matter if it takes a while: assuming cargo isn't spending all their money on mothroaches or gambling crates, giving them even 100k worth of diamonds completely removes the need for them to do bounties or mail for the rest of the shift, turning cargotechs into passengers with extra access. Which in turn removes work for some other departments (service gets no bounties to cook/grow stuff, for example)

I'm strictly of the opinion that Salvage should be Salvage, not an Infinite Money Printer

Barheru:
> The other day logi made over 1 million by selling a crate of diamonds. Even 3/u might be too much.

Rune/Donovan (who posted that screenshot) claimed that the big money in that image came from the other excess materials as well (provided you're talking about the recent screenshot in genchat that you commented on). It might be worth looking into seeing what ratio of that was diamonds versus the other materials pictured. There weren't just diamonds in that sale, and it's unclear from that image what the ratio of that over one million haul was.

Baximusu:
> > The other day logi made over 1 million by selling a crate of diamonds. Even 3/u might be too much.
> 
> Rune/Donovan (who posted that screenshot) claimed that the big money in that image came from the other excess materials as well (provided you're talking about the recent screenshot in genchat that you commented on). It might be worth looking into seeing what ratio of that was diamonds versus the other materials pictured. There weren't just diamonds in that sale, and it's unclear from that image what the ratio of that over one million haul was.

In my experience, it is mostly the sale of gold, silver, goliath plates, hivelord cores and diamonds. While it is nice to get such things to get a major lead, it's becoming too common that Logistics just swims in funds after an hour, making any work sort of moot. I was speaking about this yesterday, and it's become too common where it's almost just the norm, now.

SyaoranFox (approved):
I also saw the power of diamonds. It can be too good if focussed.

Mnemotechnician:
Neat.

PR #857, opened 2026-07-24
Global Ghost Hearing Removal and Subtle Protection Fix

Cprn-Sprkl:
## About the PR

- Removes global hearing from non-admin observers. Their hearing now follows line of sight and range rules.
- Prevents non-admin observers from seeing subtle at all.

This change protects subtle messages they way they were intended to be, and also allows for the establishing of public ERP zones without blasting every ghost with raunchy messages.

## Media

https://github.com/user-attachments/assets/e736cc08-ef20-4017-b3a4-abeca2e810bc

</p></details>

## Requirements

## Licensing:

**Changelog**

:cl: sprkl
- remove: Non-admin observers no longer have global hearing.
- fix: Non-admin observers can no longer see subtle messages at any range.

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -970,12 +970,10 @@ public sealed partial class ChatSystem : SharedChatSystem
             // Floofstation edit end
 
             // even if they are a ghost hearer, in some situations we still need the range
-            if (distance < voiceGetRange)
+            if (distance < voiceGetRange && !observer)
             {
-                if (isDead && !observer)
-                    recipients.Add(player, new ICChatRecipientData(distance, observer, Subtle: false, InLOS: isVisible));
-                else
-                    recipients.Add(player, new ICChatRecipientData(distance, observer, InLOS: isVisible));
+                // Euphoria | Hide Subtle from Non-Admin Ghosts
+                recipients.Add(player, new ICChatRecipientData(distance, observer, Subtle: !isDead, InLOS: isVisible));
                 continue;
             }
 
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -961,6 +961,7 @@ public sealed partial class ChatSystem : SharedChatSystem
                 continue;
 
             var observer = ghostHearing.HasComponent(playerEntity);
+            var isDead = HasComp<GhostComponent>(playerEntity);
 
             // Floofstation edit - check LOS
             sourceCoords.TryDistance(EntityManager, transformEntity.Coordinates, out var distance);
@@ -971,7 +972,10 @@ public sealed partial class ChatSystem : SharedChatSystem
             // even if they are a ghost hearer, in some situations we still need the range
             if (distance < voiceGetRange)
             {
-                recipients.Add(player, new ICChatRecipientData(distance, observer, InLOS: isVisible));
+                if (isDead && !observer)
+                    recipients.Add(player, new ICChatRecipientData(distance, observer, Subtle: false, InLOS: isVisible));
+                else
+                    recipients.Add(player, new ICChatRecipientData(distance, observer, InLOS: isVisible));
                 continue;
             }
 
@@ -984,7 +988,7 @@ public sealed partial class ChatSystem : SharedChatSystem
     }
 
     // Floofstation: add inLOS
-    public readonly record struct ICChatRecipientData(float Range, bool Observer, bool? HideChatOverride = null, bool InLOS = true)
+    public readonly record struct ICChatRecipientData(float Range, bool Observer, bool? HideChatOverride = null, bool Subtle = true, bool InLOS = true)
     {
     }
 
--- a/Content.Server/Ghost/GhostSystem.cs
+++ b/Content.Server/Ghost/GhostSystem.cs
@@ -240,7 +240,9 @@ namespace Content.Server.Ghost
         private void OnMapInit(EntityUid uid, GhostComponent component, MapInitEvent args)
         {
             _actions.AddAction(uid, ref component.BooActionEntity, component.BooAction);
-            //_actions.AddAction(uid, ref component.ToggleGhostHearingActionEntity, component.ToggleGhostHearingAction);
+            //Euphoria | Ghost hearing toggle action for admins only
+            if (HasComp<GhostHearingComponent>(uid))
+                _actions.AddAction(uid, ref component.ToggleGhostHearingActionEntity, component.ToggleGhostHearingAction);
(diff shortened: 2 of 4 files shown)


DoubleHypercube:
I think removing global hearing from _every_ ghost is a bit much. Surely there is a better solution here? Why not just make it opt-in by default instead of opt-out?

Mudyin:
yea there was a PR for global ghost hearing and opt-in as well as a warning via a textbox, It got closed tho?

Content.Server/Chat/Systems/ChatSystem.cs
```diff
@@ -971,7 +972,10 @@ private Dictionary<ICommonSession, ICChatRecipientData> GetRecipients(EntityUid
             // even if they are a ghost hearer, in some situations we still need the range
             if (distance < voiceGetRange)
             {
-                recipients.Add(player, new ICChatRecipientData(distance, observer, InLOS: isVisible));
+                if (isDead && !observer)
+                    recipients.Add(player, new ICChatRecipientData(distance, observer, Subtle: false, InLOS: isVisible));
+                else
+                    recipients.Add(player, new ICChatRecipientData(distance, observer, InLOS: isVisible));
```
Mnemotechnician:
```suggestion
                recipients.Add(player, new ICChatRecipientData(distance, observer, Subtle: false, InLOS: isVisible, Subtle: isDead));
```

Mnemotechnician (commented):
Other than that should be fine

DoubleHypercube:
I still don't think this should be merged without a reason as to why we're flat out removing global hearing from all player ghosts. If not hearing public scenes is the issue, surely it could just be made opt-in?

Mnemotechnician:
> I still don't think this should be merged without a reason as to why we're flat out removing global hearing from all player ghosts. If not hearing public scenes is the issue, surely it could just be made opt-in?

1. There's no point in having global hearing on any server with 20+ players. Local messages, emotes, whispers flood the chat, making it unreadable.
2. There are consent issues with hearing whispers from across the map, both on the observer and on the player side.

Content.Server/Ghost/GhostSystem.cs
```diff
@@ -240,7 +240,7 @@ private void OnGhostShutdown(EntityUid uid, GhostComponent component, ComponentS
         private void OnMapInit(EntityUid uid, GhostComponent component, MapInitEvent args)
         {
             _actions.AddAction(uid, ref component.BooActionEntity, component.BooAction);
-            _actions.AddAction(uid, ref component.ToggleGhostHearingActionEntity, component.ToggleGhostHearingAction);
+            //_actions.AddAction(uid, ref component.ToggleGhostHearingActionEntity, component.ToggleGhostHearingAction);
```
Mnemotechnician: Also, you may want to add an _admin.IsAdmin check here to allow admin ghosts to still have this action.

DoubleHypercube:
Personally I like seeing global emotes/speech visible over just radio messages.
Though, I _can_ sort of see your point about whispering; though why not just exclude that as well as subtle, since I doubt people will be using regular emotes to ERP?

ScarlettJFG:
> Personally I like seeing global emotes/speech visible over just radio messages. Though, I _can_ sort of see your point about whispering; though why not just exclude that as well as subtle, since I doubt people will be using regular emotes to ERP?

I- will admit to having occasionally accidentally hit @ instead of - when doing ERP emotes...I've gotten better at making sure now but still, it does happen, so I understand hiding of Emotes too.

Although now that I think of it, isn't it line of sight based?

Content.Server/Chat/Systems/ChatSystem.cs
Cprn-Sprkl: This has the Subtle argument twice, but I see what you're going for. Going to apply it and test.

Content.Server/Ghost/GhostSystem.cs
Cprn-Sprkl: Good idea, better than admins manipulating the ghost hearing component when they want to shut things off.

Content.Server/Chat/Systems/ChatSystem.cs
Mnemotechnician: oh yea, i copied it from one of the lines.

Content.Server/Ghost/GhostSystem.cs
Cprn-Sprkl: IsAdmin doesn't seem to work here. I assume the mind hasn't been transferred to the admin ghost entity when the init that gives the actions runs. Found an alternative that works though.

Cprn-Sprkl:
Since this is a part of restoring public ERP zones on the server it is expected that people will start using regular emotes for NSFW content again.

Observers will hear all conversation/emotes around them following the same line of sight and distance rules as characters. Only subtle and subtle ooc is fully blocked from observers.

DoubleHypercube:
> following the same line of sight and distance rules as characters.

Does this account for the fact ghosts physically do not render walls and so have closer to a sphere of sight?

Mnemotechnician (approved):

Content.Server/Ghost/GhostSystem.cs
```diff
@@ -240,7 +240,9 @@ private void OnGhostShutdown(EntityUid uid, GhostComponent component, ComponentS
         private void OnMapInit(EntityUid uid, GhostComponent component, MapInitEvent args)
         {
             _actions.AddAction(uid, ref component.BooActionEntity, component.BooAction);
-            _actions.AddAction(uid, ref component.ToggleGhostHearingActionEntity, component.ToggleGhostHearingAction);
+            //Euphoria | Ghost hearing toggle action for admins only
+            if (HasComp<GhostHearingComponent>(uid))
```
Mnemotechnician, line 244: The ghost hearing component is removed when ghost hearing is toggled. This wont work. You need to get the ghosts session (from ActorComponent i think) and then call IsAdmin on it.

Mnemotechnician (commented):
ALthough

Content.Server/Ghost/GhostSystem.cs
Cprn-Sprkl: I can certainly try that in a bit, but is there a situation where it will matter?

This code is only parsed when the ghost entity is being made and admin ghost has ghost hearing by default because they have a whole separate definition yml than regular observers. In testing admin ghosts correctly get the action on creation and because they already have the action they can toggle on and off all they want.

Is there a foresee-able edge case where an admin ghost is spawning but is also somehow already stripped of the component?

Mnemotechnician: oh, i see. I guess it's not that important then.

SyaoranFox (approved):

PR #860, opened 2026-07-25
Interaction verb improvements

Mnemotechnician:
## About the PR
Improves the system by letting verbs distinguish whether they are being created based on the global pool of verbs, the user's OwnInteractionVerbs, the target's InteractionVerbs, or the used item's ToolInteractionVerbs components.

The primary goal of this pr was to make plushies huggable and pettable. But while on it, i decided to also add the ability for used items to facilitate verbs, which resulted in me adding a "point gun at target" verb.

(first screenshot is old which is why the verbs are at the top)

image

image

image

## Why / Balance
There is a new field on the InteractionVerbPrototype that determines which facilities can add it (by default most verbs are ignored if they are facilitated by the held item, just in case someone fucks up and tries to add e.g. Bite to a carp plushie as a tool verb. Which... may work? I don't know, i think i'm overreacting here).

There is also a new requirement that lets verbs to check what they are facilitated by. For example, the Hug, Pet, and Shake verbs now pass CanPerform checks if either of the following is true: the target is a mob, or this verb was facilitated by the target's InteractionVerbs component (explicitly added). This allows to add verbs like pet and hug to any entity, not just mobs.

Also, you wont fucking believe it. There are like 10 guns all inhereting from BaseItem and having no common parent.  What the fuck? I made BaseGun just to make them share a parent.

## Media

https://github.com/user-attachments/assets/dc0f6228-08b8-49f4-a157-27d49c2e6bc2

</p></details>

## Licensing:
    

## Breaking changes
GetInteractionVerbsEvent and InteractionVerbsSystem.AddAll now accept pairs of (protoid, InteractionVerbSource) instead of just protoid.

**Changelog**
:cl:
- add: You can now hug and pet plushies, and point your guns at people using new interaction verbs.
- fix: Waving at people should no longer expose the evil virtual item you shouldn't see.


Dunrab (approved):
fih

PR #862, opened 2026-07-25
Add lethal ranged to secborg escalation module. Refactor items inside secborg modules. #862

mrfluxstone:
(NOTE: This is my first PR here. Please don't be mean or I will cry)

### WHAT WAS CHANGED
- Standardize Security Equipment: All security modules now feature a whistle and sec light. These items occupy the last two slots of every module and ensure that a sec light is always available no matter which module is currently active.

- Improve the Patrol Module: Adding a crowbar to the patrol module. This gives the module a practical role during power outages, sabotage, and APC shutdowns by allowing Security Borgs to access airlocks and secure sensitive areas when normal access systems are unavailable.

- Remove improv handcuffs from de-escalation module and replace it with normal cuffs.

- Add a Lethal Ranged Option to the Escalation Module, the Silicone PDW-9: The weapon is balanced with **reduced damage, lower fire rate, and a slow recharge time**. It has no disabler and is strictly lethal. It does **20 heat damage and provides 7 shots**. It is designed to handle dangerous vent critters or fight one unarmored person. (See discussion below)

### WHY WAS IT CHANGED
Here are some arguments I picked up during talking about the “no ranged lethal options” issue with others:

**1. “Borgs should assist their department not replace actual crew”**

I understand the concerns surrounding lethal ranged weapons for Security Borgs, but I believe the proposed Silicon PDW-9 addresses them while still allowing Security Borgs to fulfill their intended role.

The weapon has intentionally limited damage, fire rate, recharge time and is locked behind a research-only module that takes some time to research. It is intended to be an option for limited covering fire or helping with dangerous vent critters, not overwhelming combat power. Taking care of dangerous threats is the job of the actual officers.

Right now it's impossible to fight Revs without a tools module which already takes up ONE of TWO available slots (the welder being the only option for heat damage). Against round ending threats they are merely cannon fodder and to some a cheap resource for healing (dragons). It cannot reasonably assist security in any way.

The goal here is that a borg cannot “win” these encounters by himself but to participate in them and provide useful support.

**2. “Sec borgs could power game with lethal weapons”**

Powergaming should not be treated as a balance argument for removing useful equipment from a role. If a player powergames, that is a rules issue. Also if you power game while playing on this server you are on the wrong server (at least in my opinion). Additionally, this module is not freely available. It requires research, and it is not exactly a high priority research option.

The module is also strictly lethal. It does not provide a disabler mode, meaning it has a clear purpose rather than being a universally superior option (we have the chase module for that)

**3. “Sec borgs should not carry -long arms- at all times”**

A reasonable concern that I would like to address in a different way. Right now the Silicon PDW-9 is to be treated as a low power sidearm like any officer can carry so I don't think this argument applies here. I do have an interesting idea for long arms though (see further discussion below).

Ranged lethal aside I did some other changes as well

1. Give the patrol module a crowbar. Because this module is baked in and its useless. Security should be able to crowbar open doors in case of massive power failure. And “wasting” a slot on a tools module just feels really bad. You could argue that this module allows the borg to enter high sensitive sites without power and guard/patrol them.
2. Remove improv handcuffs from de-escalation module and replace it with normal cuffs: Because why would a security officer carry flimsy makeshift handcuffs when they have access to the real deal?
3. Add the sec light and the whistle to every module. Some modules had them. Some did not. Sometimes you had a sec light, sometimes you did not. It was just annoying really and doesn't change much of the balancing. Though it would be cool if the security borg had a “sec light flashlight” in its eyes so we could remove these items entirely.

### Further Discussion
I was thinking about giving the Sec Tech role some more stuff to do (and to give it another purpose besides steal interactions away from the poor roboticist role):
Create a module for, let's say, a laser rifle. This module ONLY has a laser rifle in it. It is treated as an actual “long arm” so just carrying it around on green would be an SoP violation. Once a code red comes around the Sec Techs job is to “arm” and “de-arm” a sec borg.

I would also like a proper “riot control” module that is capable of using flashes, stingers and then has a home for the bean bag shotgun that is currently living inside the escalation module. I did not have the time yet to tinker with this (THIS IS MY FIRST PR GUH) and I am aware that a non lethal launcher is inside the bastion module already.

I was also thinking about giving the Sec Borg one more slot so it could actually fit some default modules or fun ones. Though right now I don't think you can distinguish like so “You can have two sec modules and one standard one”. Would be fun for the future.


mrfluxstone:
Patrol
Image
Bastion
Image
Det
Image
Chase
Image
Escalation
Image
Deescalation
Image

### Short preview can be found at https://youtu.be/KMCH0oJvlSY?si=F2l7jImltKQSXSYF

Resources/Prototypes/_DV/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
```diff
@@ -46,6 +46,53 @@
     soundHit:
       collection: MeatLaserImpact
 
+- type: entity #Projectile for the Silicon PDW-9, reduced damage for balancing reasons
```
Dunrab: same here

Resources/Prototypes/_DV/Entities/Objects/Specific/Robotics/borg_modules.yml
```diff
@@ -50,9 +50,10 @@
     - state: icon-patrol
   - type: ItemBorgModule
```
Dunrab, line 51: any changes to an upstream file need to be commented with # Euphoria or # Floofstation so we know someone in our community made the change and its not lost with a upstream merge or merge conflict

Resources/Prototypes/_DV/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
```diff
@@ -652,3 +652,55 @@
     proto: BulletRadiation
     fireCost: 100
   - type: Appearance
+
+- type: entity
```
Dunrab, line 664: all new entities need to go under the _Floof or _Euphroia namespace

Dunrab (changes requested):
a few small things youll need to fix

Resources/Prototypes/_DV/Entities/Objects/Specific/Robotics/borg_modules.yml
mrfluxstone: Added # Euphoria Tags while trying to follow the guidelines

Resources/Prototypes/_DV/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
mrfluxstone: moved to Panta-Rhei/Resources/Prototypes/_Euphoria/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml

Resources/Prototypes/_DV/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
mrfluxstone: moved to Panta-Rhei/Resources/Prototypes/_Euphoria/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml

mrfluxstone:
Feedback applied. Thanks for taking the time to read over these.

Resources/Prototypes/_DV/Entities/Objects/Specific/Robotics/borg_modules.yml
```diff
@@ -70,6 +71,7 @@
     - item: FlashRecharging
     - item: WeaponLauncherNonLethal
     - item: SecurityWhistle
+    - item: FlashlightSecliteRecharging
```
Dunrab: you added the Flashlight to this module as well so youll want to mark this with #Euhphoria to

mrfluxstone: Guh... changed. Thanks boss.

mrfluxstone:
Kindly requesting another review 👀

mrfluxstone:
Humbly requesting another review (though given the various upstream issues I would understand if there are bigger problems rn) @Dunrab

PR #865, opened 2026-07-26
Tweaks and fixes

Mnemotechnician:
## About the PR
Hairbrushes:
- Now correctly display the popups
- Now have a sound (same sound as wirebrushes, but quieter and limited in distance)
- Now have a repeating do-after

Leashes:
- Changing leash length now correctly dirties the leash (used to cause bugs with leash prediction enabled)
- Fixed a bug that triggered an ancient engine bug that caused containers containing both the leash and the leashed entities to become immovable
- When the leash is made shorter, the actual joint only changes maxLength when the entities are close enough. There was a bug preventing it from working.
- Are now predicted by default (there's a command to toggle leash prediction)

Loadouts: added a muzzle to the trinkets group (requested by @Dunrab)

## Technical details

PopupOnUse:
- Added a "repeat" field to the component
- Cleaned up the code and fixed a minor bug with audio
- Changed the popup call from PopupClient to PopupPredicted ||(for some reason the do-after event is only predicted on the side of the user? why does the system even bother serializing shit then if the only client who's going to see it already predicts the StartDoAfter call??)||

Leash system:
- Now RefreshJoints() is called on every tick instead of whenever leash/leashed are moved in/out of containers. There's no way to make the old approach work other besides making a marker component for the leash holder, which i frankly don't want to bother doing. This will increase the performance cost of the system, but not by much. If the leash system becomes a performance hog however, we will have to do that (and in process change how/when the system performs joint relay refreshes)
- For years i thought that Joint.Length is the current length of the joint. I was wrong. It's the natural length which the joint tries by exerting finite force. MaxLength and MinLength are distance ranges in which entity coordinates are clamped with infinite force. T## About the PR
Hairbrushes:
- Now correctly display the popups
- Now have a sound (same sound as wirebrushes, but quieter and limited in distance)
- Now have a repeating do-after

Leashes:
- Changing leash length now correctly dirties the leash (used to cause bugs with leash prediction enabled)
- Fixed a bug that triggered an ancient engine bug that caused containers containing both the leash and the leashed entities to become immovable
- When the leash is made shorter, the actual joint only changes maxLength when the entities are close enough. There was a bug preventing it from working.
- Are now predicted by default (there's a command to toggle leash prediction)

Loadouts: added a muzzle to the trinkets group (requested by @Dunrab)

## Technical details

PopupOnUse:
- Added a "repeat" field to the component
- Cleaned up the code and fixed a minor bug with audio
- Changed the popup call from PopupClient to PopupPredicted ||(for some reason the do-after event is only predicted on the side of the user? why does the system even bother serializing shit then if the only client who's going to see it already predicts the StartDoAfter call??)||

Leash system:
- Now RefreshJoints() is called on every tick instead of whenever leash/leashed are moved in/out of containers. There's no way to make the old approach work other besides making a marker component for the leash holder, which i frankly don't want to bother doing. This will increase the performance cost of the system, but not by much. If the leash system becomes a performance hog however, we will have to do that (and in process change how/when the system performs joint relay refreshes)
- For years i thought that Joint.Length is the current length of the joint. I was wrong. It's the natural length which the joint tries by exerting finite force. MaxLength and MinLength are distance ranges in which entity coordinates are clamped with infinite force. The only reason joint.Length doesn't matter in our case is that leash joints have zero force.

## Media

Compression go brrrrrr

https://github.com/user-attachments/assets/00a9cca3-2d33-4ef9-a492-f8b10eafc234

</p></details>

## Requirements

## Licensing:

**Changelog**
:cl:
- fix: Putting a leashed entity and the leash in the same container should no longer make the container immovable. This includes the space dragon which would sometimes get stuck due to this.
- fix: When the length of a leash is shortened, the "rope" will only get shorter when the leashed entity comes close enough - as opposed to changing instantly and forcing it to come closer. Also other players should now correctly see length changes.
- tweak: Leash prediction is now enabled by default. This should make leash physics visually better. You can still toggle this using the /leashpredict command.
- he only reason joint.Length doesn't matter in our case is that leash joints have zero force.

## Media

Compression go brrrrrr

https://github.com/user-attachments/assets/00a9cca3-2d33-4ef9-a492-f8b10eafc234

</p></details>

## Requirements

## Licensing:

**Changelog**
:cl:
- fix: Putting a leashed entity and the leash in the same container should no longer make the container immovable. This includes the space dragon which would sometimes get stuck due to this.
- fix: When the length of a leash is shortened, the "rope" will only get shorter when the leashed entity comes close enough - as opposed to changing instantly and forcing it to come closer. Also other players should now correctly see length changes.
- tweak: Leash prediction is now enabled by default. This should make leash physics visually better. You can still toggle this using the /leashpredict command.
- tweak: Hairbrushes now make sound when used, and their do-after automatically repeats.
- fix: The popup from hairbrushes should now show correctly for other players.
- tweak: Muzzles are now available as loadout trinkets.

### DIFF
--- a/Content.Shared/_Floof/CCVar/FloofCCVars.Leash.cs
+++ b/Content.Shared/_Floof/CCVar/FloofCCVars.Leash.cs
@@ -8,5 +8,5 @@ public sealed partial class FloofCCVars
     ///     Client-only, defines whether the client should predict leash joints.
     /// </summary>
     public static readonly CVarDef<bool> PredictLeashes =
-        CVarDef.Create("leash.predict", false, CVar.CLIENTONLY | CVar.ARCHIVE);
+        CVarDef.Create("leash.predict", true, CVar.CLIENTONLY | CVar.ARCHIVE);
 }
--- a/Content.Shared/_Floof/Leash/LeashSystem.ContainerWorkarounds.cs
+++ b/Content.Shared/_Floof/Leash/LeashSystem.ContainerWorkarounds.cs
@@ -7,25 +7,29 @@ public sealed partial class LeashSystem
 {
     private void InitializeContainerWorkarounds()
     {
-        SubscribeLocalEvent<LeashedComponent, EntGotInsertedIntoContainerMessage>(OnLeashedContainerChanged);
-        SubscribeLocalEvent<LeashedComponent, EntGotRemovedFromContainerMessage>(OnLeashedContainerChanged);
-
-        SubscribeLocalEvent<LeashComponent, EntGotInsertedIntoContainerMessage>(OnLeashContainerChanged);
-        SubscribeLocalEvent<LeashComponent, EntGotRemovedFromContainerMessage>(OnLeashContainerChanged);
+        // This doesn't really work because if the parent of the leash holder changes, we won't be notified
+        // I don't want to add another component just for the leash holder (lazy), so i just made these checks be performed on each tick
+        // They are pretty cheap anyway
+        //
+        // SubscribeLocalEvent<LeashedComponent, EntGotInsertedIntoContainerMessage>(OnLeashedContainerChanged);
+        // SubscribeLocalEvent<LeashedComponent, EntGotRemovedFromContainerMessage>(OnLeashedContainerChanged);
+        //
+        // SubscribeLocalEvent<LeashComponent, EntGotInsertedIntoContainerMessage>(OnLeashContainerChanged);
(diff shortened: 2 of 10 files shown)


Mnemotechnician:
I think i legit wrote more text in the pr description than changed code.

Dunrab:
> I think i legit wrote more text in the pr description than changed code.

What is this slander

Dunrab (approved):
looks fishy to me

PR #881, opened 2026-07-30
Ninja tweak: Dash do after

Cprn-Sprkl:
## About the PR

Adds a half second 'do after' to the dash ability.

## Why / Balance

There has been a long running [thread](https://discord.com/channels/1255902263667331193/1496264824550850702) on the discord discussing this. The idea of a short 'do after' to discourage use of the teleport mid-combat was arrived at.

## Technical details

Adds handling for a do after to the DashAbilitySystem.

## Media

https://github.com/user-attachments/assets/947450bc-0842-4a18-be76-598bd82382ba

</p></details>

## Requirements

## Licensing:

## Breaking changes

N/A

**Changelog**

:cl: sprkl
- tweak: Ninja dash now has a short wind up.

### DIFF
--- a/Content.Shared/Ninja/Systems/DashAbilitySystem.cs
+++ b/Content.Shared/Ninja/Systems/DashAbilitySystem.cs
@@ -1,7 +1,6 @@
-using System.Reflection.Metadata.Ecma335;
 using Content.Shared.Actions;
 using Content.Shared.Charges.Systems;
-using Content.Shared.DoAfter;
+using Content.Shared.DoAfter; // Euphoria | Dash do after
 using Content.Shared.Hands.EntitySystems;
 using Content.Shared.Interaction;
 using Content.Shared.Movement.Pulling.Components;
@@ -9,10 +8,10 @@ using Content.Shared.Movement.Pulling.Systems;
 using Content.Shared.Ninja.Components;
 using Content.Shared.Popups;
 using Content.Shared.Examine;
-using Robust.Shared.Audio;
-using Robust.Shared.Audio.Systems;
-using Robust.Shared.Map;
-using Robust.Shared.Serialization;
+using Robust.Shared.Audio; // Euphoria | Dash do after
+using Robust.Shared.Audio.Systems; // Euphoria | Dash do after
+using Robust.Shared.Map; // Euphoria | Dash do after
+using Robust.Shared.Serialization; // Euphoria | Dash do after
 
 namespace Content.Shared.Ninja.Systems;
 
--- a/Content.Shared/Ninja/Systems/DashAbilitySystem.cs
+++ b/Content.Shared/Ninja/Systems/DashAbilitySystem.cs
@@ -1,5 +1,7 @@
+using System.Reflection.Metadata.Ecma335;
 using Content.Shared.Actions;
 using Content.Shared.Charges.Systems;
+using Content.Shared.DoAfter;
 using Content.Shared.Hands.EntitySystems;
 using Content.Shared.Interaction;
 using Content.Shared.Movement.Pulling.Components;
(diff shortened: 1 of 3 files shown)


Content.Shared/Ninja/Systems/DashAbilitySystem.cs
```diff
@@ -76,6 +84,57 @@ private void OnDash(Entity<DashAbilityComponent> ent, ref DashEvent args)
             return;
         }
 
+        //Euphoria | Setup information for dash do after
+        comp.Destination = args.Target;
+        comp.Dasher = args.Performer;
+        comp.Handled = TryStartDashDoAfter(ent);
+
```
Mnemotechnician, line 88: These should be properties of the event, why the hell are they added to the component?

Content.Shared/Ninja/Systems/DashAbilitySystem.cs
```diff
@@ -76,6 +84,57 @@ private void OnDash(Entity<DashAbilityComponent> ent, ref DashEvent args)
...
+            BreakOnDamage = true,
+        });
+    }
+
+    //Euphoria | Dash do after
+    private void OnDashDoafter(Entity<DashAbilityComponent> entity, ref DashDoAfterEvent args)
+    {
+        if (args.Cancelled)
+        {
+            _sharedCharges.AddCharges(entity.Owner, 1);
```
kivdon, line 132: If I'm running around and press the "teleport" multiples times while running, thus causing it to be cancelled multiple times, is it going to grant me a new charge each time? Is it using up a charge every time I even make an attempt to use it, and so this is refunding it?

Cprn-Sprkl: Before the do after code there is a TryCharge which uses up a charge when it executes. It's before the do after so you don't have to wait to find out you have no charges. The code you quoted here is for returning that charge in the event your dash is cancelled.

Content.Shared/Ninja/Systems/DashAbilitySystem.cs
Cprn-Sprkl: Great suggestion. Everything after the comma was not needed and comes across as hostile.

Mnemotechnician (approved):

Content.Shared/Ninja/Systems/DashAbilitySystem.cs
Mnemotechnician: i see

sorry.

Content.Shared/Ninja/Systems/DashAbilitySystem.cs
```diff
@@ -1,12 +1,18 @@
+using System.Reflection.Metadata.Ecma335;
```
SyaoranFox: Please add markers to the using(s)

SyaoranFox (changes requested):
Once that is done, i'll merge it.

Content.Shared/Ninja/Systems/DashAbilitySystem.cs
Cprn-Sprkl: Done and done. I was previously told that adding comments to using section was useless because IDEs will strip them if someone auto fixes their using section.

SyaoranFox: I dunno, just want it to be as correct as possible.

DoubleHypercube: Yes and no; the comments get preserved but can be jumbled around (ordering isn't preserved).
What I'm curious about is how reflection got imported at all here- (even if it was removed afterwards)

Cprn-Sprkl: No idea. I switched from VS Code to Rider a couple of months ago and keep encountering strange things every time I do something in it.

Mnemotechnician: Yes, it is completely useless
When i do an upstream merge, i usually accept all changes to using statements and then tell rider to optimize imports (and auto-import missing things in process)

SyaoranFox (approved):

PR #888, opened 2026-07-31
Innate Thieving Component

kivdon:
## About the PR

Makes it so upon receiving the Thief role, a Thief also gets the Thieving component. 
The Thief still gets thieving gloves.
The innate Thieving component is stealthy, but has no strip time reduction. 

## Why / Balance

There have been complaints of powergaming/metagaming, with Sec taking away thief gloves when they shouldn't, or shouldn't even know what they are. 
Thieves now can remain stealthy even without the gloves, but do not get a bonus to the speed at which they steal unless they are wearing the gloves. 

## Technical details

Just yml changes.

## Media

image
The right side is the Innate Thief stealing without gloves on. 
The left side is the victim not knowing they're being stolen from. 

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- tweak: Made it so Thieves have an innate ability to steal things stealthily, even without gloves. The gloves are still needed to do it quickly, though.

### DIFF
--- a/Resources/Prototypes/GameRules/subgamemodes.yml
+++ b/Resources/Prototypes/GameRules/subgamemodes.yml
@@ -27,10 +27,10 @@
       allowNonHumans: true
       multiAntagSetting: NotExclusive
       startingGear: ThiefGear
-       components:
+      components:
       # - type: Pacified # DeltaV - forced pacifism is lame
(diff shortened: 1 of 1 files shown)


kivdon:
Tested.

Mnemotechnician (approved):

PR #895, opened 2026-08-03
Display Case Bugfix

Raikyr0:
## About the PR
Makes display cases disable the artifact inside.

## Why / Balance
Someone got swapped into an display case and when they went to break said case to escape, they got round removed.

**Changelog**
:cl:
- fix: Fixed Display Cases activating.


flutteristhebest:
Sealed away forever...

ScarlettJFG:
Thank you Raikyr!

Mnemotechnician (approved):

PR #901, opened 2026-08-04
Adds Rhino Suit to LO Locker

Raikyr0:
## About the PR
Puts the Rhino Suit into the LO Locker.

~~Enables the Golden Knuckle Dusters in LO Locker.~~ Irrelevant now upstream did that.

## Why / Balance
Both exist as roll-able steal objectives with no way to complete due to it being missing.

Also the Rhino suit technically does exist else where, but it's in the LO's SUIT locker which isn't mapped on any of our maps. I imagine it works fine on DV if their maps actually have it.

Rather then remove the objective, I felt it would be better to make use of the asset rather then let it become unused.

**Changelog**
:cl:
- add: Added LO's Rhino Suit to their locker (for steal objective reasons).


Mnemotechnician:
Why does the LO need a hardsuit with armor similar to that of the nukie hardsuits?

Raikyr0:
> Why does the LO need a hardsuit with armor similar to that of the nukie hardsuits?

Having a look at it, it's much worse then a combat one.

Nukie 0.9x Movement Speed
- type: Armor
    modifiers:
      coefficients:
        Blunt: 0.5
        Slash: 0.5
        Piercing: 0.5
        Heat: 0.5
        Radiation: 0.5
        Caustic: 0.5
        Cold: 0.8

Rhino 0.7x Movement Speed
- type: Armor
    modifiers:
      coefficients:
        Blunt: 0.7
        Slash: 0.5
        Piercing: 0.7
        Heat: 0.6
        Radiation: 0.3
        Caustic: 0.7

Goliath 0.8x Movement Speed (The one Salvage can made with a little work.)
- type: Armor
    modifiers:
      coefficients:
        Blunt: 0.7
        Slash: 0.7
        Piercing: 0.5
        Heat: 0.7
        Radiation: 0.3
        Caustic: 0.8

SyaoranFox:
> > Why does the LO need a hardsuit with armor similar to that of the nukie hardsuits?
> 
> Having a look at it, it's much worse then a combat one.
> 
> Nukie 0.9x Movement Speed
> 
> * type: Armor
>   modifiers:
>   coefficients:
>   Blunt: 0.5
>   Slash: 0.5
>   Piercing: 0.5
>   Heat: 0.5
>   Radiation: 0.5
>   Caustic: 0.5
>   Cold: 0.8
> 
> Rhino 0.7x Movement Speed
> 
> * type: Armor
>   modifiers:
>   coefficients:
>   Blunt: 0.7
>   Slash: 0.5
>   Piercing: 0.7
>   Heat: 0.6
>   Radiation: 0.3
>   Caustic: 0.7
> 
> Goliath 0.8x Movement Speed (The one Salvage can made with a little work.)
> 
> * type: Armor
>   modifiers:
>   coefficients:
>   Blunt: 0.7
>   Slash: 0.7
>   Piercing: 0.5
>   Heat: 0.7
>   Radiation: 0.3
>   Caustic: 0.8

Okay, so why does LO need that?

Arbitrated:
We could see benefits to the Logistics Officer having access to an EVA suit of their own, for operations such as rescuing salvagers, flying damaged Cargo/Salvage Shuttles, and so on. 

And in regards to rescuing salvagers, having some protection against physical threats would make sense! 

We do think that the suit might be a bit too combat-tuned for the server, likely because the suit itself was built/balanced around Nukies - Cargo/Supply/Logistics is a common entry point for Nukie Squads.

Raikyr0:
Personally, the LO needs a EVA suit due to their job nature (LO and HoP are the only heads without a suit, but the HoP doesn't need to go into space regularly nor does service get hit by meteors 3 times a shift), the suit stats itself is around the same as a Salvage suit. Also the higher slash resist only really matters against carps so people with guns or PKAs still blow em up.

M3739 (changes requested):
This PR assumes that a prototype having a `StealTarget` component is enough justification for it being included in the game. It is not. Such a component only details how it is utilized after it exists. It does not justify adding a weapon, nor a hardsuit.

Unless the premise is revised, this is not something I would like to see merged.

Raikyr0:
> This PR assumes that a prototype having a `StealTarget` component is enough justification for it being included in the game. It is not. Such a component only details how it is utilized after it exists. It does not justify adding a weapon, nor a hardsuit.
> 
> Unless the premise is revised, this is not something I would like to see merged.

It seems that upstream has enabled the knuckle dusters regardless so that point doesn't matter anymore. 
The hardsuit however is justifiable since the LO is a job that needs to go into space, and doesn't have a dedicated space suit, unlike the CMO. I will however point out, this is more of a mapping issue, since the hardsuit is in a LO hardsuit locker, but none of our maps have it mapped.

PR #906, opened 2026-08-06
Improved Pyrogenics Chems

Raikyr0:
## About the PR
Improved Pyrogenics Chems.
Requires 5 less degrees (celsius) and heals 50% more.

## Why / Balance
Pyrogenics is much harder to setup (heating a cryopod without knowing atmos will overpressure the pod and crush the person inside as well as cook them) with a much larger downside with no upside over the extremely safe cryogenics (can be under minus 200 degrees with no downsides).

This PR makes pyrogenics more viable as a higher risk to reward.

**Changelog**
:cl:
- tweak: Changed Pyrogenics chemicals to be better then Cryogenics chemicals as the risk and reward wasn't matching.


Mnemotechnician:
Heating gas won't "crush the person" unless you're heating it by burning the person or the pod is already overpressurised. An increase from ~297°K to 380°K will only raise pressure by ~30% which is not even remotely close to hurting the person given that the cryopod is filled with standard 1 atm air.

It takes 250 kpa of pressure to actually start hurting your fellow spaceman, which would require heating 1 atm air all the way to 750°K at which point your spaceman will have turned to ash.

PR #907, opened 2026-08-06
Add: Telepathic Trait

DoubleHypercube:
## About the PR

It do what it say it do

## Why / Balance

Lets you be roundstart telepathic without any of the baggage of psionics.

## Technical details

Minor injection into `ChatUIController` and Nyanotrasen's `NyanoChatSystem` to make this work.

## Media

Example Media Embed

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl: DoubleHypercube
- add: Telepathic trait!

### DIFF
--- a/Resources/Locale/en-US/_Euphoria/traits/mental.ftl
+++ b/Resources/Locale/en-US/_Euphoria/traits/mental.ftl
@@ -1,0 +1,2 @@
+trait-telepathic-name = Telepathic
+trait-telepathic-desc = Whether through a noöspheric uplink, innate psionic connection, training, or otherwise — you have the ability to telepathically communicate your thoughts.
--- a/Resources/Prototypes/_Euphoria/Traits/mental.yml
+++ b/Resources/Prototypes/_Euphoria/Traits/mental.yml
@@ -1,0 +1,10 @@
+- type: trait
+  id: TraitTelepathic
+  name: trait-telepathic-name
+  description: trait-telepathic-desc
+  category: Psionics
+  cost: 1
+  effects:
+  - !type:AddCompsEffect
+    components:
+    - type: Psionic


Mnemotechnician (changes requested):
The way you're doing it right now makes no sense. On the client side, all psionics grant telepathy, and the telepathy component also grants telepathy? But then on the server side you're only checking for telepathy and not for psionics.

You should just give the entity PsionicComponent (like all psionic traits do) and call it a day unless you have an actual reason.

DoubleHypercube:
> But then on the server side you're only checking for telepathy and not for psionics.

Checking for both feels clunky when you can just check for one and EnsureComp, no?

> You should just give the entity PsionicComponent (like all psionic traits do) and call it a day unless you have an actual reason.

IIRC (or at least, I've been told) this will prevent rolling actual psionic powers. Command gets a free reroll which is why you can take them and still be psionic in dev

ASaltySeabass:
> > But then on the server side you're only checking for telepathy and not for psionics.
> 
> Checking for both feels clunky when you can just check for one and EnsureComp, no?
> 
> > You should just give the entity PsionicComponent (like all psionic traits do) and call it a day unless you have an actual reason.
> 
> IIRC (or at least, I've been told) this will prevent rolling actual psionic powers. Command gets a free reroll which is why you can take them and still be psionic in dev

I have dispel as a trait and sometimes roll random psionics on top of it. I don't play command either.

DoubleHypercube:
Huh, interesting. That settles that then. I guess the claim that it prevents it is wrong.

Mnemotechnician:
> IIRC (or at least, I've been told) this will prevent rolling actual psionic powers. Command gets a free reroll which is why you can take them and still be psionic in dev

Instead of listening to what people say, you should check the code: https://github.com/Floof-Station/Panta-Rhei/blob/9eeef91b6f98509c9017ab8a5e92044caba6bb4c/Content.Shared/_DV/Psionics/Systems/SubSystems/SharedPsionicSystem.Rolling.cs#L33
The chance to roll psionics is fixed in PotentialPsionicComponent. When you spawn in or when you take drugs, the system first checks if `Rolled = false` (it is false by default and after being hit by a noospheric discharge), then rolls a chance to give you a new power. After that it sets rolled = true to prevent you from getting free drug rerolls until the next discharge.

I'm pretty sure you're also not accounting for the fact that psionics can be mindbroken which should remove telepathy.

Mnemotechnician (approved):

SyaoranFox (approved):

PR #909, opened 2026-08-06
Various fixes and tweaks

Mnemotechnician:
## About the PR
- Fixes nukie ghost roles having invalid locales
- Reworks the ConditionalHealingSystem (it was completely non-functional due to its absurd attempt to hijack the existing HealingSystem, which was later changed in a way that made that approach impossible)
- Adds super-synth to the musician loadout (in job-specific trinkets)
- Gives the mothership core LanguageKnowledge (binary, tau-ceti)
- Changes the attack sound of melee weapons given by the Talons and Claws weapons to pierce.ogg which is what clawed species use (instead of AlienClaw)
- Makes mittens replace your physical damage with stamina damage similarly to boxing gloves (this is something dunrab wanted to do but supposedly forgot). Also gives them fibers and a fingerprint mask as intended.
- Fixes another nukeops test fail caused by wizden's assumption that all species are organics
- Disables role whitelists in development builds by default cause annoying
- Removes HumanHealth from expect alert prototypes in AlertsTest because it's dumb

## Technical details
The reworked conditional healing system spawns a virtual item in your other hand, gives it HealingComponent, invokes the HealingSystem, and despawns the item when healing is done. So effectively it requires having 2 hands in order to operate. It doesn't refund the item if the do-after is cancelled however.

## Media

IPC healing (note that different cable types heal different amounts, i also fixed the popup later):

https://github.com/user-attachments/assets/42c9e3a8-9ebf-4b46-a636-85fcf9aa215f

New mittens:

https://github.com/user-attachments/assets/712a7301-9cc7-447c-9bdb-b548f716d844

Misc:
image
image

</p></details>

## Licensing:

**Changelog**
:cl:
- fix: Fixed a bunch of things. See the PR #909 for detials
- tweak: The unarmed melee granted by the Claws and Talons traits now uses the same sound as the Rodentia unarmed melee.
- tweak: Mittens now make you deal no physical damage when worn, mask your fingerprints, and leave fibers.

### DIFF
--- a/Content.IntegrationTests/Tests/GameObjects/Components/Mobs/AlertsComponentTests.cs
+++ b/Content.IntegrationTests/Tests/GameObjects/Components/Mobs/AlertsComponentTests.cs
@@ -87,7 +87,7 @@ namespace Content.IntegrationTests.Tests.GameObjects.Components.Mobs
                 Assert.That(clientAlertsUI.AlertContainer.ChildCount, Is.GreaterThanOrEqualTo(3));
                 var alertControls = clientAlertsUI.AlertContainer.Children.Select(c => (AlertControl) c);
                 var alertIDs = alertControls.Select(ac => ac.Alert.ID).ToArray();
-                var expectedIDs = new[] { "HumanHealth", "Debug1", "Debug2" };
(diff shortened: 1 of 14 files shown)


Dunrab (approved):
lgtm, this comment isnt fishy enough though

Mnemotechnician:
lfg

PR #913, opened 2026-08-07
Make the Salvage Shuttle Map Based 

Dunrab:
## About the PR
Salv shuttles can now be set per map (just like the cargo shuttle), so each map can have a different shuttle that better suits the slot count and type of map. I was originally just going to set it up for a few maps, but then decided to just get it out of the way for all of them.

## Why / Balance
The reclaimer is making its grand return.

## Technical details
There are two versions of the reclaimer, mining.yml is the completed shuttle, while there is now a mining-broken.yml that is the broken down version that needs some repairs but can be used on round start. Another day and some more yml.

If a new salvage shuttle is created, its grid needs to be given the SalvShuttleFTL tag so it can ftl to the ATS.

## Media

No to sure how i can provide media of this

</p></details>

## Requirements

## Licensing:

**Changelog**

:cl:
- add: The Salvage Reclaimer shuttle has been spotted in use again on Asterisk, Edge and Pebble.


SyaoranFox (changes requested):
Please, while you are already here, also add the component to the salvage shuttles that would allow it to FTL to the ATS.

Dunrab:
> Please, while you are already here, also add the component to the salvage shuttles that would allow it to FTL to the ATS.

Yeah i can add it to the reclaimer & pathfinder. Do we want to keep Surgeon slots open still since we no longer have shitmed?

Cprn-Sprkl:
I don't see any harm in keeping them for RP surgery since we won't be unmapping all the surgery rooms. Just a flavour role until surgery is back in.

SyaoranFox:
> > Please, while you are already here, also add the component to the salvage shuttles that would allow it to FTL to the ATS.
> 
> Yeah i can add it to the reclaimer & pathfinder. Do we want to keep Surgeon slots open still since we no longer have shitmed?

Please keep it in for now.

Mnemotechnician (commented):
All of the new station prototypes still spawn just the pathfinder in the mining spawn group

Dunrab:
> Please, while you are already here, also add the component to the salvage shuttles that would allow it to FTL to the ATS.

This parts done our current two salv shuttle versions can FTL to the ats.

SyaoranFox (approved):

Dunrab:
merge conflict should be be fixed

Mnemotechnician (approved):

PR #914, opened 2026-08-08
Loko 14 Glow-down

Raikyr0:
## About the PR
Makes Loko 14 do what it says on the can. 
"The MBO has advised crew members that consumption of Fourteen Loko may result in seizures, blindness, drunkeness, or even death. Please Drink Responsibly."

I'll probably have to test it later, 15 minutes to open dev is hot ass.

## Why / Balance
The can was falsely advertising that you would suffer. I have fixed that, you will actually suffer now. 

**Changelog**
:cl:
- tweak: Loko 14 now does what it says on the can.

### DIFF
--- a/Resources/Prototypes/Reagents/Consumable/Drink/soda.yml
+++ b/Resources/Prototypes/Reagents/Consumable/Drink/soda.yml
@@ -280,6 +280,43 @@
         damage:
           types:
             Poison: 1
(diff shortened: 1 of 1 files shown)


Mnemotechnician (approved):
This makes it much more fun to drink as a non-slime i think

SyaoranFox (approved):

PR #916, opened 2026-08-08
Kick the (emotional) supports out from under the building

DoubleHypercube:
## About the PR

Renames all emotional support collars to ID collars & removes the "Emotional Support " tag from the job title.

## Why / Balance

The collars are great! They look pretty, they free up an ID slot, and they hide the ugly PDA sprite.
_However._ The pigeonholing into being an "emotional support crewmember" is decidedly less great. (Realistically, what is this even support to mean? Is it suggesting you are equivalent to like... a dog meant to keep people happy?)
So now they're just called ID collars and use normal job titles.

## Technical details

I went with the "remove custom job titles from the ID cards" option. Alternatively I could have made them use normal ID cards, however the codebase for those is fragmented. Several ES ID cards are "`EmotionalSupport<primary>`", others don't even have "IDCard" in the _name_ (looking at you, Medical), and some are completely different (e.g. `EmotionalSupportCommandHOPIDCard`).

_Alternatively_ alternatively I could have duplicated the collars into emotional support/non-emotionalsupport variants, but that feels... kind of bloaty for little gain? It also would have had the same issue as #1, being that they would have to use normal ID cards, which means rewriting them _anyway._

There is also of course the option of letting players define a custom job title, but that would require special-cased hacky workarounds in the loadout menu and also may or may not be a good idea anyway.

## Media

Example Media Embed

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl: DoubleHypercube
- tweak: Emotional support ID collars are now just ID collars.

### DIFF
--- a/Resources/Prototypes/_Floof/Entities/Clothing/Neck/Collar_Ids/collars_ids.yml
+++ b/Resources/Prototypes/_Floof/Entities/Clothing/Neck/Collar_Ids/collars_ids.yml
@@ -1,9 +1,9 @@
-﻿# base pet id collar (also is the black collar)
+﻿# base pet ID collar (also is the black collar)
 - type: entity
   parent: PassengerPDA
   id: ClothingNeckCollarBasetId
   abstract: true
-  name: black emotional support id collar
+  name: black emotional support ID collar
   description: A cute black collar that holds a pets id.
   components:
   - type: Pda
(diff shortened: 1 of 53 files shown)


Mnemotechnician:
> Is it suggesting you are equivalent to like... a dog meant to keep people happy?)

Yes, that's exactly what they meant to mean. I agree with renaming the collars on other jobs though. It's a bit too much bloat.

The passenger one should stay as it is though.

Dunrab:
As the person who made this I heavily disagree. This was a voted on design choice not bloat.

DoubleHypercube:
I don't think the pigeonholing is the way to go here. I cannot speak for everyone in the server, but I myself do not like being forcibly labeled as "emotional support" for deigning to use the collar; and I am very, _very_ sure that others will reciprocate that opinion. There are little alternatives other than "simply duplicate everything" - which runs into the same problem as the fact that the collar IDs do not have names similar to existing IDs (and duplicating the set of items just for that would unquestionably be bloat) - or "let players specify their own custom job title", which would require loadout menu hacks.

Mnemotechnician:
The option to create alternative variants that do not have this part of the job title is still here.

DoubleHypercube:
That would require spriting visually distinct ID collars to avoid being bloat. Which runs into the dual issues of "why can't I have the nice looking collar without being branded equivalent to a dog for someone's health" and "who is going to sprite all of those?".

Mnemotechnician:
You don't need unique sprites. Simply renaming the existing ones is enough.

DoubleHypercube:
> There is also of course the option of letting players define a custom job title, but that would require special-cased hacky workarounds in the loadout menu and also may or may not be a good idea anyway.

It turns out this does _not_ require hacky workarounds!
So I implemented that instead, and will close this PR in favour of it.

PR #917, opened 2026-08-08
Aiden Gauss (Bread Terrorist) has arrived on the station!

DoubleHypercube:
## About the PR

Allows players to rename their job titles. Great for allowing more customization beyond what ΔV allowed with the PDA subtypes.

## Why / Balance

Spawned out of #916; see that PR for the primary reasoning I did this.

## Media

Example Media Embed

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl: DoubleHypercube
- add: You can now change your job title to anything!
- tweak: Emotional support collars are now just called ID collars.


Mnemotechnician (changes requested):
This field should only be shown on applicable items

DoubleHypercube:
I don't think we can do that without duplicating the XAML and controller to specialcase, can we? Do components exist at this point in setup?

Mnemotechnician:
> I don't think we can do that without duplicating the XAML and controller to specialcase, can we? Do components exist at this point in setup?

Control.Visible = false

DoubleHypercube:
Yeah that. That would work, huh.

As far as I can tell `_entMan.TryGetComponent<PdaComponent>(_previewContainer.SpriteEntity, out _);` is the only way to do this but if there's a better way (since this feels kinda hacky) I'm open to changing it.

PR #919, opened 2026-08-08
Hotfixes

Mnemotechnician:
## About the PR
- Increases the alcohol metabolism rate of animal livers to 1x (from 0.1x). The low rate of metabolism was causing reagents to have very minor effects, preventing drunkness buildup when consuming ethanol.
- Removes the replacement accent from ratking
- Makes emotes respect LOS (again)
- Fixes WaggingSystem action granting on species without default tail markings (it was listening on the wrong event: profile import instead of appearence load)
- Fixes mittens giving you a wide swing attack. Mittens now allow disarm just like unarmed melee.

## Media

image

image

</p></details>

**Changelog**
- fix: Species with animal livers should now metabolize alcohol at a normal rate.
- fix: The "toggle wagging" action should now be correctly added if you have an animate-able tail.
- fix: Ratking no longer has a replacement accent.
- fix: Emotes respect LOS again.
- fix: You can disarm/shove while wearing mittens.
- fix: IPCs can have hair and have more marking points on the head.

### DIFF
--- a/Content.Client/Weapons/Melee/MeleeWeaponSystem.cs
+++ b/Content.Client/Weapons/Melee/MeleeWeaponSystem.cs
@@ -140,7 +140,7 @@ public sealed partial class MeleeWeaponSystem : SharedMeleeWeaponSystem
         if (altDown == BoundKeyState.Down)
         {
             // If it's an unarmed attack then do a disarm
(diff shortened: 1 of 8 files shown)


Mnemotechnician:
Gonna merge this now so we can maybe update today

PR #920, opened 2026-08-08
Health Analyzer - Allergies

ShepardToTheStars:
## About the PR

Adds a list of allergies to the health scanner.

## Why / Balance

Feels like something that should be available on scan. It does not tell you how severe the allergies are though. You still still have to look that up.

~~Depending on how many allergies one has, this could be a long list, so we could possibly limit it to like 5 or 10 if needed.~~ Fixed by just putting it after damage mostly.

## Technical details

Allergies are a server-side component only, so a serialized list has to be in the payload sent from server to client. I resisted moving that all to shared though because scope creep. That looks like it could be done pretty easily to be honest.

If the health analyzer ever becomes [predicted](https://github.com/space-wizards/space-station-14/pull/42773), allergies should be moved to shared and become networked. 

## Media

### With Allergies (1)
image

### With Allergies (way too many)
image

### Without Allergies
image

</details>

## Requirements

## Licensing:

## Breaking changes

None.

**Changelog**

:cl:
- add: The health scanner now shows allergies!

### DIFF
--- a/Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml
+++ b/Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml
@@ -74,9 +74,6 @@
 
     </BoxContainer>
 
-    <PanelContainer Name="AllergiesDivider" Visible="False" StyleClasses="LowDivider" /> <!-- Euphoria - Allergies -->
-    <BoxContainer Name="AllergiesContainer" Visible="False" Orientation="Vertical" Margin="0 5" /> <!-- Euphoria - Allergies -->
-
     <PanelContainer StyleClasses="LowDivider" />
 
     <BoxContainer
@@ -85,4 +82,6 @@
         Orientation="Vertical">
     </BoxContainer>
 
+    <PanelContainer Name="AllergiesDivider" Visible="False" StyleClasses="LowDivider" /> <!-- Euphoria - Allergies -->
+    <BoxContainer Name="AllergiesContainer" Visible="False" Orientation="Vertical" Margin="0 5" /> <!-- Euphoria - Allergies -->
 </BoxContainer>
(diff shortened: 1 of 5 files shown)


DoubleHypercube:
I'm on the fence about this. On one hand, it detracts from RP and contributes to scanmed. On the _other_, though, it _is_ a useful tool for when you don't have time to check for allergies or they aren't in records.

ShepardToTheStars:
I'm not a huge fan of the character records' UI myself, which is why I PR'd this. It's very clunky.

I also considered adding the records to a device network and only showing it if the network was up and powered, but didn't because scope creep.

Mnemotechnician:
This removes basically all RP aspects from allergies.

ShepardToTheStars:
> This removes basically all RP aspects from allergies.

I disagree because needing to ask over comms or pull out your tablet and check for every patient really only adds tedium. I could add this type of information to the character records UI (better presented) but at that point, why not just make an allergy scanner.

That aside, what if I made it show the reagent group instead, so that you still have to ask over comms or look it up. But you can take a chance if its a dire situation and don't have time.

kivdon:
Would we feel better if instead of being added to the Health Analyzer, there was a separate Allergy Analyzer device instead? Could lock it behind research, and it would add to the tedium that already exists for getting allergy information, instead of removing the tedium.

ScarlettJFG:
> Would we feel better if instead of being added to the Health Analyzer, there was a separate Allergy Analyzer device instead? Could lock it behind research, and it would add to the tedium that already exists for getting allergy information, instead of removing the tedium. 

What if we did that as like a T2 with it being a "100%, these are the confirmed allergens" and made the medical scanner say that there's an allergen in "X category"? 

Like say you scan Joe Shmoe and he is allergic to THC, it would read "Allergen detected in Narcotics Category" or something similar?

I also only say T2 cause that's around the time medical usually gets mats...

SyaoranFox:
Diagnosis RP is important for medical here. If you want to bring something like this in, lock it behind research.

PR #923, opened 2026-08-10
Vampirism!

DoubleHypercube:
## About the PR

Adds a Bloodsucker trait. It does exactly what you think it does.
Removes uncooked animal proteins (UAPs) from Blood reagent as well.

## Why / Balance

A _lot_ of people want vampirism or adjacent. So here we are!

Removing UAPs was needed to stop you from vomiting and taking damage when drinking normal human blood. Adding thirst satiation and protein was to preserve the existing functionality which UAPs served and to allow obligate vampirism (this one was requested by @Aureuix). Also - yes, the health analyzer will say 110% blood or something like that; this isn't a bug and will happen if you drink blood normally. And also for some reason drinking blood reduces your blood level at the same time it increases it, but again, happens if you just chug a glass of it. SS14 shitcode at work I guess lmao.

## Technical details

Removed the old Floof BloodsuckerComponent/System/etc due to being likely non-relicenseable (said by @Dunrab)

## Media

Github has a file size limit of 10MB so have a Discord link instead
https://cdn.discordapp.com/attachments/1536083892958929006/1536159658811588638/suck.mp4?ex=6a7a63ab&is=6a79122b&hm=901032ddf10176a625812327ad4d45e280490368c486123a098473ae71761a66&

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl: DoubleHypercube
- add: New Bloodsucker trait!
- remove: Blood no longer adds uncooked animal proteins to your chemstream.
- tweak: Blood now directly restores hunger and thirst for all metabolism types.


Mnemotechnician:
We already have a bloodsucker system that was used for the HollowFangs and Vampirism traits pre-rebase, why duplicate it?

DoubleHypercube:
> We already have a bloodsucker system that was used for the HollowFangs and Vampirism traits pre-rebase, why duplicate it?

As far as I'm aware we have not been able to successfully relicense that under MIT, which is why I made this.

Mnemotechnician:
I'm not sure where you're getting this info from, but if that is the case, you should remove it.

DoubleHypercube:
It was @Dunrab who mentioned it in a thread on Discord; I was hesitant to remove it initially but I'll take that as a go-ahead.

Mnemotechnician (approved):

Resources/Prototypes/Reagents/biological.yml
```diff
@@ -19,15 +19,18 @@
...
+      #   type: [ Human ]
+      #   inverted: true
     Food:
       effects:
-        - !type:AdjustReagent
-          reagent: UncookedAnimalProteins
-          amount: 0.1
+      - !type:AdjustReagent # Euphoria: For vampirism
+        reagent: Protein
+        amount: 0.5
```
Mnemotechnician: 0.5 units of protein per unit of blood is so much it's beyond "way too much". Drinking a single mouse would be equivalent to eating 10 steaks if not more. I'd argue even 0.1 is too much.
```suggestion
        amount: 0.1
```

Mnemotechnician (changes requested):
actually

Resources/Prototypes/Reagents/biological.yml
DoubleHypercube: Fixed, though honestly I don't know a "good" value nor do I know if it's possible to simulate hunger directly so I don't need to wait for half an hour in a devenv.

Mnemotechnician (approved):

PR #942, opened 2026-08-14
Ninja: Forbidden Target

Cprn-Sprkl:
## About the PR

This PR addresses the following issues:
- Pebble had a default warp point named Evac at Evac, which was a valid ninja bomb target
- The dorms warp point was a valid ninja bomb target
- Station beacons marking places we don't want bombed were valid ninja bomb targets

## Why / Balance

Ninjas should not be asked to bomb places they aren't allowed to bomb.

## Technical details

Touches the Pebble map file (which loaded fine in testing afterwards), reparents the default warp point for evac to the no bomb verstion, and tags several beacons that would be used in no bomb locations with the bomb blocker tag.

## Media

Pictured, a ninja not getting dorms as a target:
image

</p></details>

## Requirements

## Licensing:

## Breaking changes

While not a breaking change, there is almost certainly more instances of improper warp point and beacon usage that will need to be noted as they are found.

**Changelog**

:cl: sprkl
- fix: Ninjas should no longer be asked to bomb places they shouldn't. (Ask in AHELP if unsure about your target!)

### DIFF
--- a/Resources/Maps/Floof/pebble.yml
+++ b/Resources/Maps/Floof/pebble.yml
@@ -69963,7 +69963,7 @@ entities:
       parent: 2
     - type: Fixtures
       fixtures: {}
-- proto: WarpPoint
+- proto: WarpPointNoBomb
   entities:
   - uid: 10816
     components:
--- a/Resources/Prototypes/Entities/Objects/Devices/station_beacon.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/station_beacon.yml
@@ -260,6 +260,9 @@
   components:
   - type: NavMapBeacon
     defaultText: station-beacon-medbay
+  - type: Tag # Euphoria | Parity with warp point blocklist
+    tags:
+    - NinjaBombingTargetBlocker
 
 - type: entity
   parent: DefaultStationBeaconMedical
@@ -317,6 +320,9 @@
   components:
   - type: NavMapBeacon
     defaultText: station-beacon-research-and-development
+  - type: Tag # Euphoria | Parity with warp point blocklist
+    tags:
+    - NinjaBombingTargetBlocker
 
 - type: entity
   parent: DefaultStationBeaconScience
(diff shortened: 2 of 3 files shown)


Mnemotechnician:
The test runner froze but there's a valid error in the logs.

CastitoWae:
are... bombing targets for ninjas defined by warp points or station beacons? or wa, cuz that COULD be a heads-up or guideline for mappers in case of

edit: nvm, i read it in code,

Mnemotechnician:
Why do you not want the ninja to bomb departments? Those are the places that would make the most sense to bomb and would be the most rewarding for engi and crew to fix. No one is going to care about and fix (beyond just sealing from space) a bombed zoo, kitchen, park, or tool storage - those are basically free games for the ninja.

Arbitrated:
We agree with Memno, bombing "real" targets is helpful for giving the ninja's bomb some actual impact on the round. Skipping out on dorms and Evac 100% makes sense for server goals though!



Also we see those potshots being taken at Chef, lmaooooooooooooo

Cprn-Sprkl:
I just tagged the beacons that were the same as the warp points that were already marked as NoBomb. Namely Epistemics, Medical, and I think Atmos? I can just trim those off if they're a problem though.

Though, if I'm doing that should I also change those warp points to be regular instead of NoBomb as well?

kivdon:
I think we have a rule against antags (besides Nukies) causing station destruction by loosing Tesla/Singulo. Maybe that's why Atmos is "NoBomb"?

Cprn-Sprkl:
Reverted all the beacons tags except Evac, Dorms, Singularity, and the escape pods. Warp points that were previously marked as NoBomb can be changed in a later update if wanted. It's a bit more important the Dorms protection gets in.

SyaoranFox (approved):

Mnemotechnician (approved):

PR #955, opened 2026-08-17
Allowing Player To Choose Whether Or Not To Randomize Character For Ghost Roles

kivdon:
## About the PR

I created a new variant of the DNA Scrambler implanter and implant. This variant works exactly the same, but is named slightly differently, and will despawn after 2 minutes. If the implanter is used, the implant itself will despawn after 2 minutes and take the "scramble DNA" action with it if not used. 

I also removed the randomized appearance (and name in some cases) of the following ghost roles, and made them spawn with my "temporary" DNA scrambler implanter in-hand:
Ninja
LPO
All Nukies that parent from SyndicateOperativeGearFullNoUplink (which I think is all of them, including Lone Op)
Syndicate Recruiter
Syndicate Arms Dealer
Synthesis Specialist
Hitman

With these changes together, this means that people spawn into the ghost roles as their normal character, and then have 2 minutes to decide whether or not they want to use the optional DNA scrambler they spawn with. If they use it, they take on a random appearance just like before this change was made. If they don't use it, it despawns so it can't be used as an antag tool. 

## Why / Balance

Many people want to play as their characters in the ghost roles. Some people do not. This optional "despawning" DNA Scrambler implant allows them to play as a randomized character instead. 
Because the implanter and implant despawn after 2 minutes, they can't be used as a tool for the antag. 

## Technical details

Many yml changes. 
I tried to make these changes to Fugitive too, but realized that with the way the GALPOL warrant paper works, a DNA Scrambler would kind of break the Fugitive. That'll have to be a different PR some day.

## Media

Here's a Lone Op spawning with my DNA Scrambler in-hand:
image

Here the implanter has been used, and the DNA Scramble action is visible. Also visible is the counter before it despawns:
image

Here's the action and implanter being gone after the time has elapsed:
image

Here's the result if I do use the "scramble DNA" action (it's as permanent as the normal version):
image

</p></details>

## Requirements

## Licensing:

## Breaking changes

Not really a breaking change, but when you spawn with an implanter in your hand, the hand icon text says it's empty. But if you examine it, it's not empty. If you drop it and pick it up again, it no longer says empty. 

**Changelog**

:cl:
- add: Added a DNA Scrambler implanter/implant variant that will despawn if not used. within 2 minutes. 
- tweak: Changed many ghost roles (Ninja, LPO, all Nukies (including LoneOp), Syndie Recruit, Syndie Arms Dealer, Synthesis Specialist, Hitman) to not have randomized appearances, and spawn with my optional "despawning" DNA Scrambler implanter in-hand. You can choose whether to be your character, or a use the scrambler to be a randomized character. You have 2 minutes to decide before the implanter/implant despawns.


Mnemotechnician:
Why `speciesOverride: Human`?

kivdon:
> Why `speciesOverride: Human`?

That's what LoneOp did in the example you showed me. I thought that would make it so it defaults to a human, but allows any species that's not on the blacklist to override it. And I have nothing on the blacklist, so any species would override it.

Mnemotechnician (changes requested):
all new prototypes need to be moved under our namespace

Resources/Prototypes/_Euphoria/Entities/Objects/Misc/implanters.yml
```diff
@@ -0,0 +1,10 @@
+- type: entity
+  id: DnaScramblerImplanterDespawning
+  name: optional DNA scrambler implanter (despawning)
```
Mnemotechnician:
```suggestion
  name: optional DNA scrambler implanter
  suffix: 120s despawn
```

Mnemotechnician (approved):

PR #964, opened 2026-08-20
Fix leashes making disposals collidable

Mnemotechnician:
## About the PR
DisposalPipeSysetm tries to move the entity into a pipe -> calls ContainerSystem.Insert -> that one calls RecursivelyUpdateJoints -> that one sets up a joint relay from the leashed entity onto the pipe -> the joint relay calls WakeBoody -> that FOR SOME FUCKASS REASON calls SetCanCollide(true) (THERES ZERO REASON TO DO THIS BS, WHY RT?!), which makes the pipe collidable

CollideOnAnchorSystem doesn't listen on CollisionChangeEvent and so cannot fix this error

All of these 3 systems are part of RT (there's literally no reason for CollideOnAnchor to be part of RT) so I had to make a new system to fill in the niche. Luckily this system should cover more possible edge cases than just leashes (e.g. jumping into disposals while attached to a grappling hook would cause the same issue)

## Media

After like 5 rounds of testing and getting the error, this solution finally works in the same scenario:

https://github.com/user-attachments/assets/6a490084-da8d-4906-93b0-9962e30cd9cd

</p></details>

## Requirements

## Licensing:

**Changelog**
:cl:
- fix: Disposal pipes should no longer become collidable when underground

### DIFF
--- a/Content.Shared/_Floof/Leash/LeashSystem.Joints.cs
+++ b/Content.Shared/_Floof/Leash/LeashSystem.Joints.cs
@@ -1,6 +1,7 @@
 using Content.Shared._Floof.Leash.Components;
 using Robust.Shared.Containers;
 using Robust.Shared.Physics;
+using Robust.Shared.Physics.Components;
 using Robust.Shared.Physics.Dynamics.Joints;
 using Robust.Shared.Physics.Systems;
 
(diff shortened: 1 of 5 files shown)


Arbitrated:
Wait a sec, is this why dragging crates around is sometimes a pain in the butt, because the disposal pipes become collidable?

SyaoranFox (approved):
I feel like we should make a PR for RT about this..

Mnemotechnician:
It seems to be fixed upstream. just a week away.

PR #968, opened 2026-08-20
Space dragon buff

XavyBoio:
## About the PR
 Rebalanced the Space Dragon to be a more liable threat

## Why / Balance
 Added damage resistances to the dragon, and added night vision based off my experiences in the past about complaints of dragons dying in maints due to having no vision. Also changed damage values of their attacks and the projectile speed and health pool, and overall blood pool. A general buff across the board.

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- add: Evolved Dragon
- remove: Bad dragon
- tweak: Updated Dragon values
- fix: Security.

### DIFF
--- a/Resources/Prototypes/Damage/modifier_sets.yml
+++ b/Resources/Prototypes/Damage/modifier_sets.yml
@@ -334,3 +334,19 @@
   id: ManifestedSpirit
   coefficients:
     Holy: 2
+
+- type: damageModifierSet #Euph - Added Damage Modifiers for Dragon Protection.
+  id: Dragon
+  coefficients:
+    Blunt: 0.75
+    Slash: 0.65
+    Piercing: 0.80
+    Heat: 0.25
+    Shock: 0.75
+    Cold: 0.50
+    Caustic: 0.75
+    Poison: 0.25
+    Radiation: 0.50
+    Asphyxiation: 0
+    Bloodloss: 1.0
+    Cellular: 0.75
--- a/Resources/Prototypes/Entities/Mobs/Player/dragon.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/dragon.yml
@@ -10,7 +10,7 @@
     bloodReferenceSolution:
       reagents:
       - ReagentId: Blood
-        Quantity: 650
+        Quantity: 700 # Euph - Increased dragons blood pool due to bleeding easily.
   - type: GhostRole
     allowMovement: true
     allowSpeech: true
@@ -38,6 +38,9 @@
     - Dragon
   - type: Speech
     speechVerb: LargeMob
+  - type: NightVision # Euph - Dragons have night vision now, rejoice.
+    drawOverlay: true
+    color: "#FF0000"
   - type: CombatMode
   - type: MobMover
   - type: InputMover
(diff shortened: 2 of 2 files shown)


XavyBoio:
Screenshot 2026-08-19 191121


Adding this here cuz spooky

Mnemotechnician:
The dragon is not supposed to act like a tank... Its purpose is to summon fishes which are the real threat in a dragon encounter.

XavyBoio:
This PR was made due to witnessing officers soloing dragons on their own with a lecter and instances of dragons having Carp or no Carp mixed in, but my experiences are based off of what I commented in my PR. 

If there is any recommendation of things to change, or deny the PR. Feel free to let me know.

Arbitrated:
The night vision is certainly nice, though that red is pretty harsh on the eyes in our opinion. 

We are concerned about the dragon potentially wiping out a station when dealing with lower-population stations (especially lower Sec population). Note that the Dragon is supposed to heal whenever it eats things, IIRC it gets Ichor added to its bloodstream after consuming a person - so making it more resistant to damage (especially in terms of resistances instead of simply More Health) can have unintended effects. 

We do wonder if having more carp could be interesting, making the Dragon a bit more "swarmy" with the carp commanding aspects. 


We love the idea of making antagonistic forces more potent, though!

Mnemotechnician:
I'm not sure. In my experience, if the dragon can accumulate ~10 fishes before the officers show up, and the dragon doesn't let its fishes get blatantly shot down in long halls (instead fighting in tight spots and exploiting choke points), it can easily overpower security, especially so during low pop.

Dunrab:
> I'm not sure. In my experience, if the dragon can accumulate ~10 fishes before the officers show up, and the dragon doesn't let its fishes get blatantly shot down in long halls (instead fighting in tight spots and exploiting choke points), it can easily overpower security, especially so during low pop.

my thoughts are if we buff the dragons we have to increase the min players required for it to spawn to save lop a bit from a buffed dragon

kivdon:
The night-vision I agree with, increasing how much blood it has is fine (though it may exacerbate the issue a lot of people have with the dragon in that it's basically encouraged to do one attack, and then go run away into space to find another angle, and just play hit-and-run like a ninja).

But I definitely can't agree with all of the quite extensive buffs you're giving it. I can tell you put a lot of thought into it, though. The dragon is one of our most common antags, and I fear that this would make it far too powerful. I already see it essentially "win" fairly often.

As well, there have been people talking about reducing the number of Sec slots, to overall "buff" all antags. These changes would be even more overtuned in that regard. 

One of the things that makes the dragon so powerful is that it can completely and entirely remove threats to it. If it eats someone, it not only heals, it prevents that person from getting back into the fight. Every time it eats a Sec officer, the station loses ability to fight the dragon. A buffed dragon could exacerbate the spiraling dragon crisis. 

Overall, I'm worried about making every dragon into a round-ending crisis.

(Also, while I appreciate the humor in the change log, it should probably be edited to be more clear about what's actually happening with it. The current text tells the average person absolutely nothing at all.)

Ray3424:
The night vision thing is good, I'm surprised it doesn't already have it since spacedragon spends a lot of time in maintenance.

The rest, I agree with everybody else about.

Spacedragons, on many servers, are intended to be reasonably probable round-ending events, and on those servers they do that job pretty well if nobody knows how to play against them and pretty decently even if they do. 

On a server like this one, where the rounds are deliberately on the longer side and aren't really intended to be abruptly ended in the same way, I think of spacedragons as less what they are on servers where they actually win semi-routinely and more like Rat King+. Intended to a problem, with a very outside chance to end the round, but probably a very solvable problem.

I don't think they need to be more than Rat King+, and I think they fill that niche just fine.

Arbitrated:
What if the Dragon started with some healing/armor that protected them until they got their first rift active, or something like that? 

Since that would mean Dragon has some extra protection at its most vulnerable (when it first arrives at the station), but it would be harder to steamroll everyone.

PR #969, opened 2026-08-21
adds jaunter slots to salvage webbing variants

badlanduser:
## About the PR

jaunter belt slots for merc webbing and military webbing salvage variants
## Why / Balance

qol for salvage, if they decide to use it
## Technical details

minor yml and ftl changes
## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- tweak: Jaunter slot added to salvage webbing


Mnemotechnician:
Test fail is valid, also i feel like adding ItemSlots to storage items is not a good idea

badlanduser:
> Test fail is valid, also i feel like adding ItemSlots to storage items is not a good idea

for the test fail, i honestly have no clue what that means.

as for the second part, we already have ItemSlots on the sec belt for a firearm and radio

Mnemotechnician:
The test is failing because ItemSlotsSystem adds a container slot on CompInit (it should only do this on MapInit because otherwise mapping an item with ItemSlots will result in unnecessary data being saved into the map file, bloating it)

You have two potential solutions:
1. Fix the bug in ItemSlotsSystem
2. Override the ContainerManager component and specify all containers the entity should have in the prototype

Mnemotechnician (approved):

SyaoranFox (approved):

PR #971, opened 2026-08-21
Update Rules, Remove Random Romerol from puddles and toilets

SyaoranFox:
## About the PR

Removed rule 7.1 as it was left in by accident, and remove random spawns of romerol. This also includes preventing ANY reagent from generating with the oracle.

## Why / Balance
Romerol is counter against the kind of gameplay we have here.

## Technical details
Its bad. I remove the allReagents filter and replace it with "RewardReagents". It essentially means you always get a reward reagent.

## Media
<details> <summary><h3>Click to show</h3></summary> <p>

Example Media Embed

</p></details>

## Requirements

- [X] I have tested all added content and changes.
- [X] I have added media to this PR or it does not require an ingame showcase.

## Licensing:

- [X] I confirm that I am the creator of the content in this PR, and allow licensing it under the following license(s), or that the original author has given me permission to do so:
  - [X] MIT (https://github.com/Floof-Station/Panta-Rhei/blob/master/LICENSE-MIT.txt) 
    

## Breaking changes

**Changelog**

:cl:
- remove: Romerol is no longer a fill option for random toilets and puddle spawns.
- remove: Removed rule 7.1. It was left in by accident, having meant to be replaced by 7.2

PR #975, opened 2026-08-21
Oracle can no longer select Romerol

deltaSpawn0040:
## About the PR

Makes the Oracle's ability to generate random reagents no longer able to generate Romerol specifically.

## Why / Balance

Can't have Romerol.

But specifically, I was told by someone that they didn't like the quick and dirty fix that's going to be done, because it doesn't allow the Oracle to generate any chemicals but those in `RewardReagents`, so no vodka or watermelon juice or such. So here's a different quick and dirty fix to consider as an alternative to that. It might be better, it might be worse. Y'all can be the judge.

## Technical details

It's a very simple and stupid change. It adds an additional requirement to the pattern-matching statement feeding into `allReagents` telling it not to select prototypes with the ID "Romerol". I checked the resulting list and it seems to work, and I think with how pattern matching works and how often it happens, the performance impact of the additional check should be negligible. That's just a guess, though.

## Requirements

## Licensing:

## Breaking changes

I do not believe this could break anything.

**Changelog**

:cl:
- tweak: Oracle can no longer dispense Romerol.

### DIFF
--- a/Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
+++ b/Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
@@ -209,7 +209,7 @@ public sealed class OracleSystem : EntitySystem
             return;
 
         var allReagents = _prototypeManager.EnumeratePrototypes<ReagentPrototype>()
-            .Where(x => !x.Abstract)
+            .Where(x => !x.Abstract && x.ID != "Romerol") //Euphoria - We can't have Romerol be a possibility
             .Select(x => x.ID).ToList();
 
         var amount = 20 + _random.Next(1, 30) + _glimmerSystem.Glimmer / 10f;


Dunrab:
this is a duplicate of #971

SyaoranFox:
> this is a duplicate of #971

Its a better fix for whats in #971 I will be removing that part from #971, and bring this in its place.

SyaoranFox (approved):

Mnemotechnician:
this is a bad way to do it (the proper way would be to make a reagent blacklist), but w/e

deltaSpawn0040:
> this is a bad way to do it (the proper way would be to make a reagent blacklist), but w/e

I'm keeping it in the back of my mind to investigate when I finish other things. `VentClogRule.cs` does something similar (and by something similar I mean the exact same block of code, character for character) and there's a comment mentioning something implementing something like that being on the list, so I might check up there if they got anything on that. I don't have high hopes, though, that comment was added in 2022.
```
        // TODO: "safe random" for chems. Right now this includes admin chemicals.
        var allReagents = PrototypeManager.EnumeratePrototypes<ReagentPrototype>()
            .Where(x => !x.Abstract)
            .Select(x => new ProtoId<ReagentPrototype>(x.ID)).ToList();
```

PR #991, opened 2026-08-25
Jump Ability Trait

Dunrab:
## About the PR
I just found out from mnemo we have had this for a while and uhm i wanted it as a trait so here we are.

## Why / Balance
Makes the jump ability into a trait you can take.

## Technical details
Some yamls.

## Media

image

https://github.com/user-attachments/assets/91f495df-630e-4e64-bed7-b4c951bc9f26

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- add: New Pounce trait has been added.

### DIFF
--- a/Resources/Prototypes/_Floof/Traits/physical.yml
+++ b/Resources/Prototypes/_Floof/Traits/physical.yml
@@ -647,3 +647,18 @@
   - !type:AddCompsEffect
     components:
     - type: Snoring
+
+- type: trait
+  id: JumpAbility
+  name: trait-jump-ability-name
+  description: trait-jump-ability-description
+  category: Skills
+  cost: 2
+  effects:
+  - !type:AddCompsEffect
+    components:
+    - type: JumpAbility
+      jumpThrowSpeed: 8 # make it a little slower
+      jumpDistance: 3
+      canCollide: true # you may be pouncing but its funny watching poeple flop over
+      action: ActionPounce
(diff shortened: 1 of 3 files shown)


Resources/Prototypes/_Floof/Traits/physical.yml
```diff
@@ -647,3 +647,18 @@
   - !type:AddCompsEffect
     components:
     - type: Snoring
+
+- type: trait
+  id: JumpAbility
+  name: trait-jump-ability-name
+  description: trait-jump-ability-description
+  category: Skills
+  cost: 2
```
Mnemotechnician: 2 points is a bit too low i think
```suggestion
  cost: 4
```

Mnemotechnician (approved):
Jumpy

Resources/Prototypes/_Floof/Traits/physical.yml
Dunrab: sounds good to me

kivdon: I should point out that with "CanCollide" turned on, this ability doesn't confer much gameplay advantage. If you collide with anything (walls, doors, people, etc), you are knocked prone for a few seconds, and drop anything you were holding. The distance is also relatively low, covering less than half the width of the ATAG, and is barely faster than sprinting for about 1 second, with a cooldown of about 10 seconds. 
I will be surprised if people make effective use of it for combat or something, since any mistake stuns you and drops your weapons. It's very easy to stun yourself in hallways with this. 
Given all that, I personally think 2 makes more sense, if not being a little high.

Mnemotechnician: 2 points is very little for any positive trait.

SyaoranFox (approved):
You spelling mistake will be here for all time.

PR #1012, opened 2026-08-30
E-Sword Research

Raikyr0:
## About the PR
This PR adds a Tier 3 Arsenal Research for Energy Swords.

## Why / Balance
Energy Swords are rather weak in DV's fork. (IE, only reflects energy bolts, advanced trudgeon does way more damage)
The recipes added in this PR require diamonds which give more use for diamonds (other then selling) and makes constructing them fairly expensive.

**Changelog**
:cl:
- add: Added Energy Sword research.

### DIFF
--- a/Resources/Locale/en-US/research/technologies.ftl
+++ b/Resources/Locale/en-US/research/technologies.ftl
@@ -41,6 +41,7 @@ research-technology-advanced-shuttle-weapon = Advanced Shuttle Weapons
 research-technology-thermal-weaponry = Thermal Weaponry
 research-technology-dual-wielding-technology = Dual Wielding Technology
 research-technology-extended-pistol-magazines = Extended Pistol Magazines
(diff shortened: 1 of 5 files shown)


Mnemotechnician:
Yeah no...

Raikyr0:
> Yeah no...

is it really that bad? E-blades are really mid.

SyaoranFox:
I am closing this PR because having e-blades come into the hands of standard crew would imbalance the scales quite heavily.

Mnemotechnician:
Not only that but you also decided to give security the ability to print the same swords that traitors have to buy for a limited currency. Considering resources on stations are basically unlimited, that's just free swords.

Raikyr0:
Technically, with an emag you can also emag a lathe and also print unlimited e-swords.
The PR's idea is to give diamonds more use then just selling for a bajillion dollars.

PR #1022, opened 2026-09-02
Neuralyzer port

thefrostbite11037:
## Why / Balance
The neuralyzer (or, as I'd like to call it here, the Mesmerizer) is a hypnosis-related tool from HardLight that people have expressed interest in seeing ported to Euphoria. This provides a mechanical backbone to hypnosis-related RP and related erotic play through a physical device (the 'Mesmerizer') and a trait ('Brainwasher') that can create, configure, and apply 'compulsions' to consenting characters that are similar in action to borg laws.

This system is in no way meant to affect the balance or avenues for conflict within the rounds in regards to the station VS antagonists - it is purely for erotic play, and this is emphasized in the pop-up that any hypnotised player will see.

## Technical details
This PR includes the full brainwashing system implementation from HardLight, with component aspects spanning Shared, Client, and Server. The ported C# code and xaml UI touches upon its own newly-added files as per the diff, as well as having minor inclusions to the traits and consent files. Alongside the ported code, there are some necessary fixes included due to the HL code conflicting with Euphoria's clothing-equip DoAfters - as such, the brainwashing is relegated to a queue overriding the Update function within BrainwasherSystem.cs.

## Media

Mesmerizer:

image
image
image
image
image
image

Hypnotism trait:
image
image
image
image
image

Consent:
image

Vendor:
image

Author Permission:
image
image

</p></details>

## Requirements

## Licensing:

## Breaking changes
Should be none...? This is my first time making a Euphoria PR though, so please please PLEASE let me know if I need to make any changes :)

**Changelog**

:cl:
- 2/09/2026: First version of the PR, full port with some slightly nice placeholder sprites, working implementation


Content.Client/Content.Client.csproj
```diff
@@ -22,9 +22,15 @@
...
   <ItemGroup>
+    <Folder Include="_HL\Brainwashing\HypnotizedPopup\" />
     <Folder Include="_EE\FootPrint\" />
     <Folder Include="_Goobstation\Power\" />
   </ItemGroup>
+  <ItemGroup>
+    <Compile Update="_DV\RoundEnd\NoEorgPopup.xaml.cs">
+      <DependentUpon>HypnotizedPopup.xaml</DependentUpon>
+    </Compile>
+  </ItemGroup>
```
Mnemotechnician, line 33: Changes to csproj files need to be reverted

Mnemotechnician (commented):
The rest is fine i guess, although code-wise it's terrible

SyaoranFox (changes requested):
Currently, I can open multiple windows examining the compulsions implanted by the hypnosis, meaning that I can just fill the screen with the exact same window. There should be a check to open/close the window like with cyborg laws.

There also should be some kind of mechanism to "shake off" your hypnosis. While yes you can just ignore the compulsions given, it still marks the person as "hypnotised".

Content.Client/Content.Client.csproj
thefrostbite11037: Mind if I ask about why? I was mainly just porting all the things from HL that worked with the system, so an explanation on what this does and why it needs to be reverted would help my understanding as a developer :) thank you!

Also yes agreed the code is kinda bad, but it _is_ kinda just ported wholesale... Could this lead to maintainability issues? Would you like it to be refactored a little? I'd need some guidance on how to do so under best practices, but I wouldn't mind TOO much if you deemed it absolutely necessary...?

thefrostbite11037:
> Currently, I can open multiple windows examining the compulsions implanted by the hypnosis, meaning that I can just fill the screen with the exact same window. There should be a check to open/close the window like with cyborg laws.
> 
> There also should be some kind of mechanism to "shake off" your hypnosis. While yes you can just ignore the compulsions given, it still marks the person as "hypnotised".

More than turning your consent off and on again? I did have plans in the followup 'bonus features' commit to add a tool for completely removing compulsions from someone (aka, what is essentially a neuralyzer with empty compulsions) as a tool for the psychiatrist, potentially. Do you think it would be better in terms of consent for this to always be a player-accessible function without outside assistance, even knowing that they do not have to go along with the hypnosis? Interested to hear your take.

Also, good catch on the window issue, thank you.

ScarlettJFG:
Test Fail doesn't look related, can we get it reran?

Content.Client/Content.Client.csproj
Notarin: The csproj change is redundant. It's a change the LSP snuck in that shouldn't have made it in. It absolutely should be dropped. Though it causes no active issues, just junk.

PR #1037, opened 2026-09-10
remove construction constraints on ALL atmos devices

BasedUser:
## About the PR
This PR *removes* all restrictions enforced on queueing and building additional atmos devices via the construction menu, bringing players up to mapper parity when Atmos is involved.

## Why / Balance
Device stacking has always been a divisive topic in the community, and it was possible for as long as Wizden's construction ghosts and systems could overlap across multiple players or even latencies. Since the former is a hot topic for exploitation and the RPD makes it seem that the former is happening *despite* Funkystation (our RPD's **original** fork - holy *shit* there is so much baggage) originally allowing device stacking to the extent the RPD permitted, I am removing the barrier and letting Engineering run wild.

Euphoria isn't a gameplay-heavy fork, so this is hopefully a reasonable enough fix to delete the gray area entirely. It also simplifies admin work.

## Technical details
<details><p>
PipeRestrictOverlapSystem (regular) is a serverside system that unanchors every atmos pipe whose one of 12 directions (effectively) conflict with another anchored pipe. This system governs ***all*** placement methods across *every* overlap-restricted entity on the same tile and was added by Wizden so that two pipe nodes cannot have the same layer on the same side in an overlapping configuration. Starlight's version of this system (did we seriously merge "EntitiySystems" as a folder) is a rewrite, but as of now it's a nonfunctioning dud. Both would theoretically work and *have* on Wizden, Starlight *and* Funky, Wizden code notably allowing cursed concoctions to exist only insofar as they didn't have common layers, or were placed by a force-anchoring condition like being Entity Spawn menu-created.

However, NoUnstackableInTile was built **far** earlier when atmos was single-layer - at the time, the restriction made absolute sense because the only *productive* way to have an amount of devices on the same layer was to have one of them be entirely obscured, going in the opposite direction, the likes. The most correct solution would be to rework it to respect PipeRestrictOverlapSystem's context but that's where *upstream* would actually want to weigh in more. I trust both Starlight and Wizden *much* more in their cohesion, so the simplest solution is to simply make our atmos ignore stacking as a shared primitive until either of the two much more coherent forks makes up their mind on how to redesign the entire system.  
Similarly, having the *RPD* enforce NoUnstackableInTile is not a direction I'm interested with because it's not going to be a complete constraint. Remember, that condition *still falters* under the condition multiple players, the Entity Spawn menu, or literally anything else spawning an anchored gas pump decide to do it in a way the overlap restriction system *already declares legal.* I don't want to see any more complaints that "X fork's atmos is better" when the issue was "a gajillion communities defined their exploit line differently."</p></details>

This PR does not affect subframe designs.

## Requirements

## Licensing:
- [X] I confirm I am the creator of the content in this PR, and allow licensing it under the following license:
  - [X] CC0-1.0 (https://creativecommons.org/publicdomain/zero/1.0/)

## Breaking changes
N/A

**Changelog**
:cl: router
- fix: Fixed the long-standing issue of RPD placement never aligning with construction. Atmos devices can now be placed anywhere, whether they stick depends on whether you're already using one of their layers or not.

### DIFF
--- a/Resources/Prototypes/Recipes/Construction/utilities.yml
+++ b/Resources/Prototypes/Recipes/Construction/utilities.yml
@@ -537,7 +537,8 @@
   placementMode: AlignAtmosPipeLayers # Funky
   canBuildInImpassable: false
   conditions:
-  - !type:NoUnstackableInTile
+  # Euphoria edit - RPD consistency
+  # - !type:NoUnstackableInTile
   alternativePrototypes: # Funky
   - GasVentPump
   - GasVentPumpAlt1
@@ -552,7 +553,8 @@
   placementMode: AlignAtmosPipeLayers # Funky
   canBuildInImpassable: false
   conditions:
-  - !type:NoUnstackableInTile
+  # Euphoria edit
+  # - !type:NoUnstackableInTile
   alternativePrototypes: # Funky
   - GasPassiveVent
   - GasPassiveVentAlt1
@@ -567,7 +569,8 @@
   placementMode: AlignAtmosPipeLayers # Funky
   canBuildInImpassable: false
   conditions:
-  - !type:NoUnstackableInTile
+  # Euphoria edit
+  # - !type:NoUnstackableInTile
   alternativePrototypes: # Funky
   - GasVentScrubber
   - GasVentScrubberAlt1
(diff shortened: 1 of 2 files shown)


Mnemotechnician:
Atmos device stacking allows a ton of exploits and doesn't really benefit anyone other than the occasional atmos powergamer trying to reinvent garteg or make frezon on 6 tiles. I don't think "we aren't a gameplay-heavy fork" and "an ai slop system allows this" are good arguments towards enabling what is essentially an infinite source of exploits.

BasedUser:
ok can you then show a solution simultaneously:
- that doesn't keep the admin treadmill
- doesn't require review effort for yourself
- makes it very clear what is and is not good atmos design

because i am tired of the exploit line

BasedUser:
this PR is officially an admin issue

PR #1041, opened 2026-09-11
Minor upstream merge fixes

Mnemotechnician:
## About the PR
- Fixes a bug in the humanoid profile system that caused all entities to have an empty string as their species representation
- Removes a "organ.Body = body" debug assert in the lewd organ system because as it turns out, this field doesn't get assigned until after OrganGotInsertedEvent is fully raised (which is dumb but w/e, i'm not messing with the upstream body system)
- Changes the names of two methods in the lewd organ system as they were highly misleading

## Media

image

image

</p></details>

**Changelog**
:cl:
- fix: Fixed species names not displaying.


M3739 (approved):
LGTM.

SyaoranFox (approved):

PR #1043, opened 2026-09-11
Preening Trait

Dunrab:
## About the PR
Someone asked for it in a thread and it took three seconds to make so why not.

## Why / Balance
Cool hats

## Technical details
I only prevented Avali from taking it because they have this already by default, we can disable more species if we want but i chose not to for customization.

## Media

I made a real specimen when testing this out on a harpy so enjoy.
image

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**

:cl:
- add: A preening trait is now selectable that will allow other species besides Avali to preen their feathers.

### DIFF
--- a/Resources/Locale/en-US/_Floof/traits/physical_traits.ftl
+++ b/Resources/Locale/en-US/_Floof/traits/physical_traits.ftl
@@ -152,3 +152,6 @@ trait-slightly-clumsy-description = You have a nack for accidentally stubbing yo
 
 trait-very-clumsy-name = Very Clumsy
 trait-very-clumsy-description = You have poor hand eye coordination and often mess up climbing over things, shooting weaponry and using electronics....
+
+trait-very-preening-name = Preening
+trait-very-preening-description = Sometimes you have feathers that you need to just remove.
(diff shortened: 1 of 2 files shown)


Mnemotechnician (approved):
Would make more sense to limit it to feathered species or give some of them different feather prototypes but w/e

M3739 (commented):
Curious, does this carry over the forensic aspects, color-matching, and the adrenaline rush feature?

M3739 (changes requested):
After looking into the code of the `Preenable` component, it seems that it would retain the adrenaline boost upon losing a feather due to blunt damage. In addition, feathers can't be cleaned. So it's not wholly a vanity trait. 

With that in mind, **I'd like for this trait to be restricted to Harpies, Resomi, and Allulalo** for the time being. Why those 3 in particular is due to the original design premise of the feather preening; it gives a unique aspect and mechanism to aid in scampering away for species known for being more fragile than their peers, while being thematically suitable for the avian/raptor species.

Consider it field-testing. If it works well, we can expand further upon both the trait and it's spread across the species later on.

M3739 (approved):
LGTM!

PR #1064, opened 2026-09-16
Orderable carpets

Eagle0600:
## About the PR
Added orderable carpet crates, one bulk carpet crate that comes with up to four different colours, and one crate with two packets of metamorphic carpet that can be used to select any colour.

## Why / Balance
The uniform printer is considered a semi-secure item since it can print actual uniforms, but is also the only way to get carpets right now. This would provide a way for crew to acquire carpet without involving any command roles.

A potential concern is that this would provide the currently-cheapest way to acquire cloth in bulk, via ordering bulk carpet and recycling it for half a stack of cloth per stack of carpet, or 60 cloth per crate. This is due to the fact that the textiles crate includes 30 durathread, which is by far the majority of its price.

## Technical details
Includes a new transmutable component, system, and bound UI. Does not modify any existing systems or components.

The system and bound UI both take fields from the component to function.

The bound user interface sends a message to the system when a prototype is selected.

The system optionally starts a do-after or goes straight to the transmutation, depending on configured delay.

The system creates a new entity, removes the original entity from any container it was in, and puts the new entity in the original entity's place/container. It then moves any mind from the original entity to the new entity, if applicable. It will then create copies of any components configured to be kept (as per the component) and apply them to the new entity, overriding the original component if necessary. Finally, the original entity is deleted.

## Media
orders

https://github.com/user-attachments/assets/79f8c6ce-3710-4f1e-9359-cf23f268005d

</p></details>

## Requirements

## Licensing:

**Changelog**
:cl:
- add: Two types of carpet crates orderable from logistics.


Oshmoo:
not that i'm against this being a solution, but i feel like moving carpet recipes to the cutter machine -- which *aren't* secure machines -- would be a far more accessible way of getting crew access to carpets

Eagle0600:
Cutter machines aren't usually available roundstart, but that's not a *bad* solution.

Oshmoo:
logi also usually don't have the funds for orders like carpets on roundstart, and the components for making a cutter machine (about 8 steel, 5 glass, 1 plastic) are far easier to get a hold of, and does not require you to constantly wait for logi to bring over modular carpets

all that being said, the transmutation components are seriously impressive work. i'd love to see this implemented for that alone

Barheru:
As a selfish Head of Personnel with paperwork for requesting carpets: can they stay in the current machine, too? That way, my horrible paperwork can incentivize building the cutter machine.

Content.Shared/_Euphoria/Item/Transmute/TransmutableSystem.cs
```diff
@@ -0,0 +1,166 @@
...
+    /// <summary>
+    /// Start transmuting a transmutable entity into another entity, respecting delay. Plays sound if applicable.
+    /// </summary>
+    private void StartTransmuteEntity(Entity<TransmutableComponent> transmutable, EntProtoId prototype, EntityUid user)
+    {
+        var comp = transmutable.Comp;
+
+        // Not working for some reason but I can't be assed.
+        if (comp.PlayTransmuteSound)
+            _audio.PlayPredicted(comp.TransmuteSound, transmutable, user);
```
Mnemotechnician: The reason this is "not working" is because for some reason this code branch is only ever executed on the server side, and server-side PlayPredicted plays the sound for everyone except the user. Simplest workaround is to add a `_net.IsServer` check and use the non-predicted variant of Play()

Content.Shared/_Euphoria/Item/Transmute/TransmutableSystem.cs
```diff
@@ -0,0 +1,164 @@
...
+
+                keptComponents.Add(_serialization.CreateCopy(component, notNullableOverride: true));
+            }
+        }
+
+        var newEntity = ReplaceEntity(transmutable.Owner, prototype);
+
+        foreach (var component in keptComponents) {
+            _entity.AddComponent(newEntity, component, overwrite: true);
+        }
```
Mnemotechnician: You should use `SharedCloningSystem.CloneComponents(transmutable, newEntity, <new cloning config prototype>)` instead of... whatever ai slop this is.

Content.Shared/_Euphoria/Item/Transmute/TransmutableSystem.cs
```diff
@@ -0,0 +1,166 @@
...
+        var doAfterEvent = new TransmutableDoAfterEvent(prototype);
+        var doAfterArgs = new DoAfterArgs(EntityManager, user, comp.Delay, doAfterEvent, transmutable, used: transmutable)
+        {
+            BreakOnDamage = true,
+            BreakOnHandChange = true,
+            BreakOnMove = true,
+            BreakOnWeightlessMove = false,
+            AttemptFrequency = AttemptFrequency.EveryTick,
+            CancelDuplicate = false,
+            BlockDuplicate = false
```
Mnemotechnician: Not cancelling duplicates => someone can start two do-afters at once, the result of the first do-after will replace the target entity and cause the second to fail... at best. At worst, if both doafters finish at the same tick, the deferred deletion of the target will allow the second transmutation of the same entity.
```suggestion
```

Content.Shared/_Euphoria/Item/Transmute/TransmutableComponent.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// List of prototypes that the item can be transmuted into
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public HashSet<EntProtoId> AvailablePrototypes { get; set; } = new();
+
+    /// <summary>
+    /// List of components to keep when transmuting
+    /// </summary>
+    [DataField(customTypeSerializer: typeof(CustomArraySerializer<string, ComponentNameSerializer>))]
+    public string[]? KeepComponents;
```
Mnemotechnician: Refer to the comment about using the cloning system.

Content.Shared/_Euphoria/Item/Transmute/TransmutableSystem.cs
```diff
@@ -0,0 +1,166 @@
...
+    {
+        TransmuteEntity(transmutable, args.Prototype);
+    }
+
+    private void OnBoundUserInterfaceMessage(Entity<TransmutableComponent> transmutable, ref TransmutableBoundUserInterfaceMessage args)
+    {
+        var user = GetEntity(args.User);
+        if (!transmutable.Comp.AvailablePrototypes.Contains(args.Prototype.Id) ||
+            !_prototype.HasIndex(args.Prototype) ||
+            !_interaction.InRangeAndAccessible(user, transmutable.Owner))
```
Mnemotechnician: This check is redundant. The user interface system closes the BUI if it's inaccessible.

Resources/Prototypes/_Euphoria/Entities/Objects/Misc/carpets.yml
```diff
@@ -0,0 +1,144 @@
+
+- type: entity
+  parent: BaseItem
+  id: FloorCarpetItemMetamorphic
+  name: metamorphic carpet
```
Mnemotechnician: The current name suggests that it's a singular carpet
```suggestion
  name: metamorphic carpet packet
```

Content.Shared/_Euphoria/Item/Transmute/TransmutableSystem.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        _doAfter.TryStartDoAfter(doAfterArgs);
+    }
+
+    /// <summary>
+    /// Transmutes the given entity into a new entity of the given prototype. Transfers components from the original to the new entity based on the TransmutableComponent whitelist.
+    /// </summary>
+    private void TransmuteEntity(Entity<TransmutableComponent> transmutable, EntProtoId prototype)
+    {
+        var componentsToKeep = transmutable.Comp.KeepComponents;
```
Mnemotechnician: As per above, you must make sure the transmutable entity is not already queued for deletion to avoid duping.

Content.Shared/_Euphoria/Item/Transmute/TransmutableSystem.cs
```diff
@@ -0,0 +1,166 @@
...
+    [Dependency] private readonly SharedDoAfterSystem _doAfter = default!;
+    [Dependency] private readonly EntityManager _entity = default!;
+    [Dependency] private readonly SharedInteractionSystem _interaction = default!;
+    [Dependency] private readonly SharedMindSystem _mind = default!;
+    [Dependency] private readonly IPrototypeManager _prototype = default!;
+    [Dependency] private readonly ISerializationManager _serialization = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
```
Mnemotechnician: Redundant
```suggestion
```

Mnemotechnician (changes requested):
Overall not too bad, but there are some obvious signs of ai-generated code. I don't know whether it was used to generate the entire system or just parts of it (like the UNUSED implementation of component cloning), so I'm not gonna outright close this PR. However, the issues with it need to be fixed.

Eagle0600:
> Overall not too bad, but there are some obvious signs of ai-generated code. I don't know whether it was used to generate the entire system or just parts of it (like the UNUSED implementation of component cloning), so I'm not gonna outright close this PR. However, the issues with it need to be fixed.

You've accused me of using AI generated code before and you were wrong before, too. I modelled most of this off existing code in the various entity systems to learn how they worked, and tweaked them until they had the desired effects. I write every line of code myself. Withdraw this accusation.

Content.Shared/_Euphoria/Item/Transmute/TransmutableSystem.cs
Eagle0600: That's consistent with what I've found, but I don't know why this is only run on the server. I'll make that change.

Content.Shared/_Euphoria/Item/Transmute/TransmutableSystem.cs
Eagle0600: I am applying the principle of not trusting client input here. A modified client could send the message arbitrarily, I'm just doing a check to see if it makes sense so the server doesn't blindly mimic it.

Mnemotechnician: This particular message is sent locally from the server-side user interface system (BUI.SendMessage wraps the event into a BoundUIWrapMessage before sending it, and the interface system receives and validates it before relaying, making sure the sender session has a ui open that could send this message). It's not that big of a deal though.

Mnemotechnician (approved):
I suppose, after further analysis, it's probable that you used some existing low-quality code for reference rather than used ai. The BUI class is clearly based on the RCD bui. Doesn't explain a couple things, but I don't want to continue bothering.

Content.Shared/_Euphoria/Item/Transmute/TransmutableSystem.cs
Eagle0600: Thank you for clarifying. Since it's truly redundant I'll remove the InRangeAndAccessible check. That should also future-proof against any additional ways to access the UI remotely, as unlikely as that is.

PR #1070, opened 2026-09-18
Micro Edit to Log File Size on any Publishes.

Myzumi:
## About the PR

I do not know any ruling about touching Files in the Tools folder but,
A Micro edit to the Tools/publish_multi_request.py file to simply Log the file size in Github Actions.
Sometimes helps with finding issues...

## Why / Balance

Publish failed because the webserver's max upload was simply smaller than the Publish files that where uploaded,
Instead of searching logs on the Host, could just check the webserver's config now and check with the action log if it fits.

## Media
None.

</p></details>

## Requirements

## Licensing:

## Breaking changes

**Changelog**


M3739 (approved):
LGTM.

PR #1083, opened 2026-09-20
Rodentia tail changes are back

SyaoranFox:
## About the PR

Rodentia can now change their base tail

## Why / Balance
Its how it was

## Requirements

## Licensing:

## Breaking changes
This brings in a lot of tails into markings that are just not tails. In theory you could end up with a floating bell.

**Changelog**

:cl:
- fix: Rodentia can change their base tail again


Mnemotechnician (approved):
weird

